PR #10, opened 2014-10-16
Fix alias matching

DenisIzmaylov:


sokra:
resolve.alias is working as expected.

DenisIzmaylov:
Thanks!

PR #15, opened 2015-05-04
DirectoryDescriptionFilePlugin: recursively look into arrays to find strings pointing to files.

Cellule:
This change is motivated by the way some bower module are described.
Sometimes they are written like
```js
{
  main: ["./file1.js", "./lib/file2.js"]
}
```
This change will look into the array and add all strings to the main modules.

[Review on Reviewable](https://reviewable.io/reviews/webpack/enhanced-resolve/15)

### DIFF
--- a/lib/DirectoryDescriptionFilePlugin.js
+++ b/lib/DirectoryDescriptionFilePlugin.js
@@ -34,6 +34,16 @@ DirectoryDescriptionFilePlugin.prototype.apply = function(resolver) {
 				return callback(e);
 			}
 			var mainModules = [];
+			function addToModule(value) {
+				if(Array.isArray(value)) {
+					for(var i = 0; i < value.length; i++) {
+						addToModule(value[i]);
+					}
+				} else if(typeof value === "string") {
+					mainModules.push(value);
+				}
+				// ignore everything else
+			}
 			for(var i = 0; i < fields.length; i++) {
 				if(Array.isArray(fields[i])) {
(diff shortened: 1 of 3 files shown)


mathieumg:
@sokra I know you said in Gitter (around the time of this PR) that instead of this fix, we should favor https://github.com/lpiepiora/bower-webpack-plugin

I'm asking you to reconsider your stance on this. This fix is already written and more and more Bower packages I've seen around use the `main` field as an array. The major problem with the plugin above is that it ignores the `resolve.modulesDirectories` configuration, and even though I have set it to `["node_modules", "bower_components"]`, it first looks for modules in `bower_components`. 

Adding to that problem, is the fact that Bower installs dependencies as siblings. Let's say for example I want to use a singleton library like `jQuery` and I have it set as an npm dependency, if one of my Bower components has it as a dependency, it's gonna load that jQuery and not the one from npm, inducing a possible version discrepancy and even just break if something uses the version from npm at the same time. (if it's set globally to be `$` using the `DefinePlugin`, for example)

I prefer npm over Bower by lengths. but some libraries are just not available through npm (and some neither through Bower, but npm complains if no `package.json` is present while Bower doesn't if there's no `bower.json` and you just provite a git url)

Thank you!

Cellule:
This is getting really old and I don't even know if the fix is still good.
It seems we don't want to add this behavior anyway so I will close this PR.
Thanks

PR #21, opened 2015-07-17
normalize filenames to nfc when readdir

disjukr:
to resolve https://github.com/webpack/webpack/pull/1227


sokra:
Thanks

sokra:
Would be great if you could add a testcase for this behavior. Either in enhanced-resolve or in webpack or both. So we don't break later..

PR #24, opened 2015-11-17
Added tests for chained symlinks

sheerun:
It turns out because the function is recursive on symlinks, readlink
is enough (normally I would use realpath).

Maybe it's little slower, but that of the scope of this PR.


sokra:
Thanks

PR #26, opened 2016-02-04
Include query when caching webpack-bug #1940

alexFaunt:
Including the query in the cacheId means it will not return if the queries are different, I believe fixing the issue here: https://github.com/webpack/webpack/issues/1940


dtothefp:
@sokra @alexFaunt I can confirm this seems to fix the issues I reported in https://github.com/webpack/webpack/issues/1940

sokra:
Thanks

PR #28, opened 2016-04-08
Restore `resolve.root` Resolver

wooliet:
Use the provided `options.root` path(s) to add `ModuleInRootPlugin`
instance(s) to the `plugins` collection. This allows for the webpack
configuration to include an absolute path or collection of paths for
the `resolve.root` property.

https://webpack.github.io/docs/configuration.html#resolve-root


wooliet:
This is an attempt to resolve [webpack issue 2291](https://github.com/webpack/webpack/issues/2291).

wooliet:
webpack 2 uses `modules` only. 
https://gist.github.com/sokra/27b24881210b56bbaff7#resolving-options

PR #22, opened 2015-09-10
Query to aliased resource with custom loader 

rpora:
Hi!

to resolve https://github.com/webpack/webpack/issues/1289

PR #33, opened 2016-07-07
support alias with a query

chinesedfan:
Enhanced #22.

Fixes  [#1289](https://github.com/webpack/webpack/issues/1289), [#1513](https://github.com/webpack/webpack/issues/1513), [#2465](https://github.com/webpack/webpack/issues/2465) in webpack.

### DIFF
--- a/lib/ParsePlugin.js
+++ b/lib/ParsePlugin.js
@@ -34,6 +34,7 @@ ParsePlugin.prototype.apply = function(resolver) {
 		}
 		return querystring.parse(query);
 	}
+
 	function stringifyQuery(obj) {
 		var tokens = [];
 		for(var key in obj) {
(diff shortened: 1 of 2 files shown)


chinesedfan:
I don't know why this pr does not trigger the CI at first. So I close it and reopen.

usirin:
Any update on this?

sokra:
I cannot merge this, because it assumes that querystrings are always key-value pairs separated with `&`, but in fact it's up to the loader implementation to choose the format. They can use any string. And this is done in the wild.

chinesedfan:
@sokra As loaders can use any string as the query string, is it impossible to implement this feature in fact?

sokra:
yep. But you can write a custom resolver plugin (fork of the alias plugin), which implement this only for loaders that use this kind of querystrings

PR #39, opened 2016-08-23
Aligning default with webpack/lib/WebpackOptionsDefaulter.js

sebastiandeutsch:
I have aligned the defaults (used web/browser where I was unsure). Next step would be to update the docs PR. What do you think?


sokra:
The default in enhanced-resolve are fine. enhanced-resolve is not a package for resolving for a browser environment. It's a general purpose resolving lib. The defaults are the node.js resolving defaults.

webpack uses it for a browser environment, which means it's webpacks job to configure it for browser.

sebastiandeutsch:
Ok got it.

PR #48, opened 2016-11-21
Fix undefined parsed.module or parsed.directory issue

jussikinnula:
When upgrading to latest Webpack, I got some errors related to parsed being undefined, so this fix is a tentative one.


sokra:
Thanks

PR #51, opened 2016-12-08
fix(alias): allow alias to contain query

TheLarkInn:
* allows one to use alias with query and it be used
* if request has query, it will be used over alias query

example 
```js
alias: {
  "d": "d?someQuery"
}
```


lib/AliasPlugin.js
```diff
@@ -20,15 +20,31 @@ AliasPlugin.prototype.apply = function(resolver) {
...
 	var onlyModule = this.onlyModule;
+	var nameQuery;
+	var aliasQuery;
+
 	resolver.plugin(this.source, function(request, callback) {
 		var innerRequest = getInnerRequest(resolver, request);
 		if(!innerRequest) return callback();
 		if((!onlyModule && innerRequest.indexOf(name + "/") === 0) || innerRequest === name) {
 			if(innerRequest.indexOf(alias + "/") !== 0 && innerRequest != alias) {
+				nameQuery = name.indexOf("?") > -1 ? name.substr(name.indexOf("?"), name.length) : undefined;
```
wardpeet: Might be better to rewrite like
```
const nameQueryPos = name.indexOf('?');
const aliasQueryPos = alias.indexOf('?');
if (nameQueryPos > -1) {
    nameQuery = name.substr(nameQueryPos, name.length);
    name = name.substr(0, nameQueryPos);
}

if (aliasQueryPos > -1) {
    aliasQuery = alias.substr(aliasQueryPos, alias.length);
    alias = alias.substr(0, aliasQueryPos);
}
```

lib/AliasPlugin.js
```diff
@@ -20,15 +20,31 @@ AliasPlugin.prototype.apply = function(resolver) {
...
+				if(aliasQuery)
+					alias = alias.substr(0, alias.indexOf("?"));
+
 				var newRequestStr = alias + innerRequest.substr(name.length);
 				var obj = assign({}, request, {
-					request: newRequestStr
+					request: newRequestStr,
 				});
+
+				if(typeof nameQuery !== 'undefined' || typeof aliasQuery !== 'undefined') {
```
wardpeet: specific reason why you would check for undefined? and not just falsy?

jamesopti:
Can I confirm that this issue is what is causing the below behavior to not work?

I have a custom loader defined in my webpack config:

Webpack Version 2.4.1

```
resolveLoader: {
    alias: {
      'myLoader': 'my-loader',
    }
}
```

And when I use it like this, webpack throws an error:

```
require('myLoader?test=true!../src/someFile.js');
```

Error:

```
ERROR in ./grunt/loaders/myLoader.js?test=true!./src/someFile.js
Module build failed:
 @ ./src/js/bundles/core/main.js 10:0-72
```

TheLarkInn:
@sokra What was the reason we left this unmerged. This was back just before we shipped v2.

PR #53, opened 2016-12-15
Fix `_readJsonSync` in CachedInputFileSystem

daltones:
It was just missing actually return the parsed JSON content.

It was affecting the reading of description files (`package.json`), leaving the content as `undefined`. Like this:
```
descriptionFilePath: '[...]/node_modules/react-router/package.json',
descriptionFileData: undefined,
descriptionFileRoot: '[...]/node_modules/react-router',
```

This PR also fixes https://github.com/benmosher/eslint-plugin-import/issues/688#issuecomment-267343188

### DIFF
--- a/lib/CachedInputFileSystem.js
+++ b/lib/CachedInputFileSystem.js
@@ -197,6 +197,7 @@ function CachedInputFileSystem(fileSystem, duration) {
 		this._readJsonSync = function(path) {
 			var buffer = this.readFileSync(path);
 			var data = JSON.parse(buffer.toString("utf-8"));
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Would you be able to add a test for this, so we can ensure this never happens again?

daltones:
@SpaceK33z That would be nice, but unfortunately I don't really know where to start.

I don't have a good understanding about the code and I saw there's a lot of things going on in that File System interface.

sokra:
Thanks, I added a test case for you...

PR #54, opened 2016-12-19
Updated ResolverFactory.js to remove unused requires

### LINKED ISSUE #50: CachedInputFileSystem is imported but never used in ResolverFactory
Is that right? There seems to be no purpose to this line:

```
var CachedInputFileSystem = require("./CachedInputFileSystem");
```

https://github.com/webpack/enhanced-resolve/blob/master/lib/ResolverFactory.js#L11

johnnyreilly:
Resolves #50 - pretty minor I know but every little helps I guess? 😄

### DIFF
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -6,9 +6,7 @@ var assign = require("object-assign");
 
 var Resolver = require("./Resolver");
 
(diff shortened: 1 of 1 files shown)


sokra:
Thanks

PR #63, opened 2017-01-25
feat(automatic-release): Generation of automatic release

wardpeet:
- added release:create to create a release with standard-version
- added release:publish to publish a release to github and npm


package.json
```diff
@@ -42,7 +43,9 @@
     "test": "mocha --full-trace --check-leaks",
     "precover": "npm run lint && npm run beautify-lint",
     "cover": "istanbul cover node_modules/mocha/bin/_mocha",
-    "travis": "npm run cover -- --report lcovonly"
+    "travis": "npm run cover -- --report lcovonly",
+    "release:create": "standard-version -i CHANGELOG.md -m \"Creating release %s\"",
+    "release:publish": "git push --follow-tags origin master; npm publish"
```
d3viant0ne: @wardpeet - Per the [published documentation](https://webpack.js.org/development/release-process/) we only need a single task `"release": "standard-version"`

From that we can chain commands on to publish different targets. You also don't need to specify the changelog.md nor is the message really necessary.

The publish shouldn't be done as an npm script, it eats publishing errors. We also don't always publish to `dist-tag`. 

In theory using an npm script could be viable for pushing tags to origin but in reality, I can copy and paste the printed command `git push --follow-tags origin master` the library spits out faster than I can type `npm run release:publish`

d3viant0ne: Also worth noting that we want to eventually automate these, the command execution needs to do a single task to allow finite script control without bloating the scripts with what would end up being duplicate functionality.

sokra:
That's not something I want to have. Sorry. Thanks for the work anyway.

PR #65, opened 2017-01-27
ci(Travis): update to webpack standard build

d3viant0ne:
_Updated Functionality_ 
- adds build matrix
 - matrix allows validations against multiple webpack versions
 - travis only executes against master
 - tests on latest release of supported node versions
 - uses current codecov push syntax

_Removed Functionality_ 
 - removes coveralls org uses codecov
 - updates eslint to current 3.14.1
 - removes js-beautify ( we have dropped that everywhere else ) //cc @sokra `yes/no`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,29 @@
 sudo: false
 language: node_js
-node_js:
-  - "4.0"
-  - "5.0"
-  - node
-script: npm run travis
-
+branches:
+  only:
+    - master
+matrix:
+  fast_finish: true
+  include:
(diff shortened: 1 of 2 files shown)


TheLarkInn:
Merge conflicts now. Might want to double check what is getting run against `eslint` I removed the `test` folder in master from eslint checks because the node plugin does _not_ like the odd behavior of requires in the fixtures (ideally we could ignore fixtures)

d3viant0ne:
All tidied up

PR #67, opened 2017-01-30
Es6 refactor use file plugin

rachel-ftw:


lib/UseFilePlugin.js
```diff
@@ -1,23 +1,29 @@
...
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-function UseFilePlugin(source, filename, target) {
-	this.source = source;
-	this.filename = filename;
-	this.target = target;
-}
-module.exports = UseFilePlugin;
+const assign = require("object-assign");
```
TheLarkInn: Ooo use Object.assign() now instead of `assign()`. Same syntax and etc. But we can get rid of this tiny package once all refactored out.

lib/UseFilePlugin.js
```diff
@@ -1,23 +1,29 @@
...
+	apply(resolver) {
+		const filename = this.filename;
+		const target = this.target;
+		resolver.plugin(this.source, (request, callback) => {
+			const filePath = resolver.join(request.path, filename);
+			const obj = assign({}, request, {
+				path: filePath,
+				relativePath: request.relativePath && resolver.join(request.relativePath, filename)
+			});
+			resolver.doResolve(target, obj, "using path: " + filePath, callback);
```
TheLarkInn: Let's do a template string here. `using path ${}`

lib/UseFilePlugin.js
rachel-ftw: √, I'll work thought this

lib/UseFilePlugin.js
rachel-ftw: √

lib/UseFilePlugin.js
TheLarkInn: Sure, for now just remove from this file for this PR, and then do the rest in a separate PR.

lib/UseFilePlugin.js
```diff
@@ -1,23 +1,27 @@
+"use strict";
```
sokra: use strict after comment

ooflorent:

PR #68, opened 2017-01-30
Es6 refactor unsafe cache plugin

rachel-ftw:


lib/UnsafeCachePlugin.js
```diff
@@ -1,41 +1,41 @@
...
-		resolver.doResolve(target, request, null, createInnerCallback(function(err, result) {
-			if(err) return callback(err);
-			if(result) return callback(null, cache[cacheId] = result);
-			callback();
-		}, callback));
-	});
-};
+module.exports = UnsafeCachePlugin;
+
+const getCacheId = request => JSON.stringify({
```
TheLarkInn: @sokra is this a named function you rely on for debugging? Otherwise I'm fine with this going anon (i guess chrome binds to the identifier's name anyways for stack debugging as of later versions of Chrome.

sokra: Function name is great for profiling. But I also think this works this way too.

lib/UnsafeCachePlugin.js
```diff
@@ -1,41 +1,41 @@
+"use strict";
```
sokra: use strict after comment

lib/UnsafeCachePlugin.js
mikesherov: http://2ality.com/2015/09/function-names-es6.html

ooflorent:

PR #69, opened 2017-01-30
Es6 refactor try next plugin

rachel-ftw:


lib/TryNextPlugin.js
```diff
@@ -1,18 +1,21 @@
+"use strict";
```
sokra: use strict after comment

ooflorent:

PR #70, opened 2017-01-30
Es6 refactor sync async file system decorator

rachel-ftw:
Line 6 of this is confusing...
Linter says `fs` isn't used, but it visably is... but when commented out tests still pass. I left it commented out.
Not sure what to do with this.


rachel-ftw:
deleted the line library that referenced the file system... 
@TheLarkInn would this work for me bc of my version of node? ...? Not wanting to fuck it up, but not knowing the details on this.

TheLarkInn:
interesting on the diff it doesn't even show that line being there, maybe was deleted from when originally was merged

lib/SyncAsyncFileSystemDecorator.js
```diff
@@ -1,11 +1,12 @@
+"use strict";
```
sokra: use strict after comment

ooflorent:

PR #71, opened 2017-01-30
Es6 refactor result plugin

rachel-ftw:


ooflorent:

PR #72, opened 2017-01-30
Es6 refactor resolver factory

rachel-ftw:
used let a bunch for those 'or' statements...


lib/ResolverFactory.js
```diff
@@ -1,97 +1,98 @@
...
+const ModuleAppendPlugin = require("./ModuleAppendPlugin");
+const UnsafeCachePlugin = require("./UnsafeCachePlugin");
 
 exports.createResolver = function(options) {
 
 	//// OPTIONS ////
 
 	// A list of directories to resolve modules from, can be absolute path or folder name
-	var modules = options.modules || ["node_modules"];
+	let modules = options.modules || ["node_modules"];
```
sokra: I guess most of these `let`s could be `const`s

lib/ResolverFactory.js
```diff
@@ -1,97 +1,98 @@
+"use strict";
```
sokra: use strict after comment

lib/ResolverFactory.js
```diff
@@ -1,97 +1,98 @@
...
 		return true;
 	};
 
 	// The file system which should be used
-	var fileSystem = options.fileSystem;
+	const fileSystem = options.fileSystem;
 
-	// Use only the sync variants of the file system calls
-	var useSyncFileSystemCalls = options.useSyncFileSystemCalls;
+	// Use only the sync constiants of the file system calls
```
sokra: typo

ooflorent:

PR #74, opened 2017-02-02
Es6 refactor node js input file system

rachel-ftw:


lib/NodeJsInputFileSystem.js
```diff
@@ -1,29 +1,48 @@
+"use strict";
```
sokra: use script after comment

lib/NodeJsInputFileSystem.js
```diff
@@ -1,29 +1,48 @@
...
+
+	readdir(path, callback) {
+		fs.readdir(path, (err, files) => {
+			callback(err, files && files.map((file) =>
+				file.normalize ? file.normalize("NFC") : file
+			));
+		});
+	}
+
+	get readFile() {
```
sokra: inefficient. assign to prototype instead.

ooflorent:
Thanks. Done in #132.

PR #77, opened 2017-02-06
Es6 refactor next plugin

rachel-ftw:


lib/NextPlugin.js
```diff
@@ -1,16 +1,18 @@
+"use strict";
```
sokra: use script after comment

ooflorent:

PR #78, opened 2017-02-07
Es6 refactor modules in hierarchic directories plugin

rachel-ftw:


lib/ModulesInHierachicDirectoriesPlugin.js
```diff
@@ -1,46 +1,47 @@
+"use strict";
```
sokra: use script should be after the license comment

lib/ModulesInHierachicDirectoriesPlugin.js
```diff
@@ -1,46 +1,47 @@
...
+		this.target = target;
+	}
+	apply(resolver) {
+		const directories = this.directories;
+		const target = this.target;
+		resolver.plugin(this.source, function(request, callback) {
+			const fs = this.fileSystem;
+			const topLevelCallback = callback;
+			const addrs = getPaths(request.path).paths.map((p) => {
+				return directories.map((d) => this.join(p, d), this);
```
sokra: second argument to `map` doesn't make sense when using arrow function

lib/ModulesInHierachicDirectoriesPlugin.js
```diff
@@ -1,46 +1,47 @@
...
+			}, this).reduce((array, p) => {
```
sokra: second argument to `map` doesn't make sense when using arrow function

ooflorent:

PR #79, opened 2017-02-09
Es6 refactor module kind plugin

rachel-ftw:


lib/ModuleKindPlugin.js
```diff
@@ -1,26 +1,29 @@
+"use strict";
```
sokra: use strict should be after the license comment

lib/ModuleKindPlugin.js
```diff
@@ -1,26 +1,29 @@
...
+		this.target = target;
+	}
+	apply(resolver) {
+		const target = this.target;
+		resolver.plugin(this.source, (request, callback) => {
+			if(!request.module) return callback();
+			const obj = assign({}, request);
+			delete obj.module;
+			resolver.doResolve(target, obj, "resolve as module", createInnerCallback((err, result) => {
+				if(arguments.length > 0) return callback(err, result);
```
sokra: `arguments` is not available in an arrow function

lib/ModuleKindPlugin.js
```diff
@@ -1,26 +1,29 @@
+"use strict";
 /*
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var createInnerCallback = require("./createInnerCallback");
+const assign = require("object-assign");
```
sokra: Use Object.assign

ooflorent:

PR #81, opened 2017-02-22
Es6 refactor main field plugin

rachel-ftw:


lib/MainFieldPlugin.js
```diff
@@ -1,48 +1,51 @@
+"use strict";
```
sokra: use strict should be after the license comment

lib/MainFieldPlugin.js
```diff
@@ -1,48 +1,51 @@
...
 /*
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var path = require("path");
+const path = require("path");
+const assign = require("object-assign");
```
sokra: Use Object.assign

ooflorent:

PR #85, opened 2017-04-20
add concord extensions dir index test

mightyiam:


mightyiam:
I was certain that this test will fail, because in my webpack setup `resolve.extensions` does not work this way. It results in the `index.js` (the second possible file by extension).

Is there any difference between using the concord extensions property and using the function API `extensions` property?

mightyiam:
Never mind. I've found my issue. It was entirely my own.

sokra:
Anyway thanks for an additional test.

PR #86, opened 2017-04-20
two tests for the extensions option

mightyiam:


mightyiam:
Again, I thought that one of these was gonna fail, but nope. So here we go.

mightyiam:
Never mind about failing. I've found my issue.

test/extensions.js
```diff
@@ -0,0 +1,35 @@
...
+	});
+	it("should resolve according to order of provided extensions (dir index)", function(done) {
+		resolver.resolve({}, fixture, "./dir", function(err, result) {
+			result.should.equal(path.resolve(fixture, "dir", "index.ts"));
+			done();
+		});
+	});
+	it("should resolve according to main field in module root", function(done) {
+		resolver.resolve({}, fixture, ".", function(err, result) {
+			result.should.equal(path.resolve(fixture, "index.js"));
```
mightyiam: So, this is the current behavior. the `main` field wins over extensions. I don't know if that's correct. It hit me hard until I noticed it.

sokra: If the main field already has an extensions, others are not used. `extensions` are only appended to the request and not exchanged.

i. e. `./abc` -> `abc`, `abc.js`, `abc.ts`.
i. e. `./abc.js` -> `abc.js`, `abc.js.js`, `abc.js.ts`

sokra:
Thanks

PR #90, opened 2017-06-15
RFC: perform resolution without logging and re-run with logging when failing.

### LINKED ISSUE #88: RFC Performance Improvement Opportunity
**TLDR:** Removing `createInnerCallback` sees a 2x performance boost.

I've been doing a lot of profiling of our webpack builds to see if there are any savings to be had. While doing this I noticed that there is a ton of garbage collection happening, and through that I found my way to this repository.

It seems that `createInnerCallback` creates many large arrays (full of strings) that go unused unless you're passing the `--verbose` flag, and your build has errors.  There may be other use cases, but that's all I've been able to find.

How can we limit the impact here?  I'd love to remove this log entirely as I don't see the value of it as an end user of webpack (webpack 3? 🤞 ), especially since disabling nets a huge performance win.  If that's not possible can we expose an option to disable?

## Charts!

 - It's a little bit difficult to highlight, but notice in this call stack that after each `loggingCallbackWrapper` there is a garbage collection event.

image

 - Here the calls to `createInnerCallback` have been removed (no more garbage collection).

image

Bottom up profile of our build before any changes:
image

After removing all the `createInnerCallback` calls:
image

Removing `createInnerCallback` calls + running with `--max-semi-space-size=1000`:
image

mikesherov:
Fixes #88 

ping @TheLarkInn @sokra would love some input cc @gdborton

### DIFF
--- a/lib/Resolver.js
+++ b/lib/Resolver.js
@@ -56,26 +56,36 @@ Resolver.prototype.resolve = function resolve(context, path, request, callback) 
 		return log.join("\n");
 	}
 
-	function onResolved(err, result) {
+	function onError(err, result) {
 		if(callback.log) {
 			for(var i = 0; i < log.length; i++)
 				callback.log(log[i]);
 		}
+
 		if(err) return callback(err);
-		if(!result) {
-			var error = new Error("Can't " + message);
-			error.details = logAsString();
-			error.missing = localMissing;
-			resolver.applyPlugins("no-resolve", obj, error);
-			return callback(error);
-		}
-		return callback(null, result.path === false ? false : result.path + (result.query || ""), result);
+
+		var error = new Error("Can't " + message);
+		error.details = logAsString();
+		error.missing = localMissing;
+		resolver.applyPlugins("no-resolve", obj, error);
+		return callback(error);
 	}
-	return this.doResolve("resolve", obj, message, createInnerCallback(onResolved, {
-		log: writeLog,
-		missing: missing,
-		stack: callback.stack
-	}, null));
+
+	function onResolve(err, result) {
+		if(!err && result) {
+			return callback(null, result.path === false ? false : result.path + (result.query || ""), result);
+		}
+
+		return resolver.doResolve("resolve", obj, message, createInnerCallback(onError, {
+			log: writeLog,
+			stack: callback.stack
+		}));
+	}
+
+	onResolve.missing = missing;
+	onResolve.stack = callback.stack;
+
+	return this.doResolve("resolve", obj, message, onResolve);
 };
 
 Resolver.prototype.doResolve = function doResolve(type, request, message, callback) {
--- a/lib/Resolver.js
+++ b/lib/Resolver.js
@@ -38,14 +38,8 @@ Resolver.prototype.resolve = function resolve(context, path, request, callback) 
 		request: request
 	};
 
-	var localMissing = [];
-	var missing = callback.missing ? {
-		push: function(item) {
-			callback.missing.push(item);
-			localMissing.push(item);
-		}
-	} : localMissing;
-	var log = [];
+	var localMissing;
+	var log;
 	var message = "resolve '" + request + "' in '" + path + "'";
 
 	function writeLog(msg) {
@@ -76,13 +70,17 @@ Resolver.prototype.resolve = function resolve(context, path, request, callback) 
 			return callback(null, result.path === false ? false : result.path + (result.query || ""), result);
 		}
 
+		localMissing = [];
+		log = [];
+
 		return resolver.doResolve("resolve", obj, message, createInnerCallback(onError, {
 			log: writeLog,
+			missing: localMissing,
 			stack: callback.stack
 		}));
 	}
 
-	onResolve.missing = missing;
+	onResolve.missing = callback.missing;
 	onResolve.stack = callback.stack;
 
 	return this.doResolve("resolve", obj, message, onResolve);
--- a/test/missing.js
+++ b/test/missing.js
@@ -1,5 +1,6 @@
 var resolve = require("../");
 var path = require("path");
+var should = require("should");
 
 describe("missing", function() {
 	var testCases = [
@@ -22,6 +23,12 @@ describe("missing", function() {
 			path.join(__dirname, "fixtures", "node_modules", "m1", "missing-file.node"),
 			path.join(__dirname, "..", "node_modules", "m1", "missing-file"),
 		]],
+		[path.join(__dirname, "fixtures"), "m1/", [
+			path.join(__dirname, "fixtures", "node_modules", "m1", "index"),
+			path.join(__dirname, "fixtures", "node_modules", "m1", "index.js"),
+			path.join(__dirname, "fixtures", "node_modules", "m1", "index.json"),
+			path.join(__dirname, "fixtures", "node_modules", "m1", "index.node"),
+		]],
 		[path.join(__dirname, "fixtures"), "m1/a", [
 			path.join(__dirname, "fixtures", "node_modules", "m1", "a")
 		]],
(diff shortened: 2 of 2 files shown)


lib/Resolver.js
```diff
@@ -78,6 +78,19 @@ Resolver.prototype.resolve = function resolve(context, path, request, callback)
 	}, null));
 };
 
+Resolver.prototype.doResolveWithErrorLogging = function doResolveWithErrorLogging(type, request, message, callback, options) {
+	var optionsWithoutLog = {
+		stack: options.stack,
+		missing: options.missing
+	};
+
+	this.doResolve(type, request, message, createInnerCallback((err, result) => {
```
gdborton: Do we need this outer `createInnerCallback` call? I know it isn't doing much, but this code path is hit a lot.

gdborton:
Seems legit to me. Thanks for picking this up!

lib/Resolver.js
mikesherov: My main concern was compatibility, it still needs to compare options.stack with callback.stack:

```
if(!log) {
		if(options.stack !== callback.stack) {
			var callbackWrapper = function callbackWrapper() {
				return callback.apply(this, arguments);
			};
			callbackWrapper.stack = options.stack;
			callbackWrapper.missing = options.missing;
			return callbackWrapper;
		}
		return callback;
	}
```

mikesherov: basically, it returns the old function after 2 conditionals if everything is OK, so it's extremely low overhead. Avoiding the GC is the huge win you discovered. I think we can afford the 2 correctness checks for now, and if it still remains a perf issue, address in a follow up... thoughts?

gdborton: That makes sense.

This is probably a micro optimization that likely has little to no impact, but I was just thinking that we wouldn't need the `callbackWrapper.stack` and `callbackWrapper.missing` with this code path anyway, since we recall `this.doResolve` w/ `createInnerCallback` on the error case.

mikesherov: Tests fail without it that it fails to throw errors with correct messages otherwise. I was on the same train of thought as you!

lib/Resolver.js
```diff
@@ -71,11 +71,24 @@ Resolver.prototype.resolve = function resolve(context, path, request, callback)
...
 		missing: missing,
 		stack: callback.stack
-	}, null));
+	});
+};
+
+Resolver.prototype.doResolveWithErrorLogging = function doResolveWithErrorLogging(type, request, message, callback, options) {
+	var optionsWithoutLog = {
+		stack: options.stack,
+		missing: options.missing
```
sokra: I case of an error, `stack` and `missing` are written twice.

lib/Resolver.js
```diff
@@ -71,11 +71,24 @@ Resolver.prototype.resolve = function resolve(context, path, request, callback)
...
-	return this.doResolve("resolve", obj, message, createInnerCallback(onResolved, {
+	return this.doResolveWithErrorLogging("resolve", obj, message, onResolved, {
 		log: writeLog,
 		missing: missing,
 		stack: callback.stack
-	}, null));
+	});
+};
+
+Resolver.prototype.doResolveWithErrorLogging = function doResolveWithErrorLogging(type, request, message, callback, options) {
```
sokra: I guess further optimization can be done if this is inlined into `resolve`. i. e. `missing` is only needed for the error run if `callback.missing` is unset.

sokra:
There are no tests for the log result yet. It's would be great to add some.

lib/AliasFieldPlugin.js
```diff
@@ -41,11 +40,11 @@ AliasFieldPlugin.prototype.apply = function(resolver) {
 			path: request.descriptionFileRoot,
 			request: data
 		});
-		resolver.doResolve(target, obj, "aliased from description file " + request.descriptionFilePath + " with mapping '" + innerRequest + "' to '" + data + "'", createInnerCallback(function(err, result) {
+		resolver.doResolveWithErrorLogging(target, obj, "aliased from description file " + request.descriptionFilePath + " with mapping '" + innerRequest + "' to '" + data + "'", function(err, result) {
```
mikesherov: `doResolve` skips the "run twice", `resolve` doesn't unless I'm misunderstanding?

lib/Resolver.js
mikesherov: Does that mean I should run without `missing` and `stack` first? Or just `stack`?

mikesherov:
@sokra I'll definitely add a bunch of tests. Just wanted to ensure I had the approach correct.

lib/AliasFieldPlugin.js
sokra: This is the control flow when using `doResolveWithErrorLogging` inside of plugins:

```
resolve -> doResolveWithErrorLogging -> SomePlugin -> doResolveWithErrorLogging -> SomeOtherPlugin -> error
                                                                          v-----------------------------+
                                                                     (with log) -> SomeOtherPlugin -> error
                               v------------------------------------------*-----------------------------+
                          (with log) -> SomePlugin -> doResolveWithErrorLogging -> SomeOtherPlugin -> error
                                                                          v-----------------------------+
                                                                     (with log) -> SomeOtherPlugin -> error
   v---------------------------*------------------------------------------*-----------------------------+
```

This is the control flow when using doResolve:

```
resolve -> doResolveWithErrorLogging -> SomePlugin -> doResolve -> SomeOtherPlugin -> error
                               v--------------------------------------------------------+
                          (with log) -> SomePlugin -> doResolve -> SomeOtherPlugin -> error
   v---------------------------*--------------------------------------------------------+
```

mikesherov: Got it.

mikesherov: So this means I should just get rid of createInnerCallback here.

mikesherov: ?

sokra: No the createInnerCallback is ok. Just leave the calls in the plugins in the original version and it should be fine.

lib/createInnerCallback.js
```diff
@@ -5,7 +5,7 @@
 module.exports = function createInnerCallback(callback, options, message, messageOptional) {
 	var log = options.log;
 	if(!log) {
-		if(options.stack !== callback.stack) {
+		if(options.stack !== callback.stack || options.missing !== callback.missing) {
```
gdborton: Is this an unrelated change?

lib/AliasFieldPlugin.js
gdborton: I'm trying to understand what's to be changed here. From the sound of it `.resolve` is the only function directly called via the build, then all the `.doResolve` calls are made from Resolver plugins... is that right?

If so it would seem to me the only thing that this PR needs to change is `.resolve` around line 74 to be something like...
```javascript
return this.doResolve("resolve", obj, message, function(err, result) {
	if (err) {
		return this.doResolve("resolve", obj, message, createInnerCallback(onResolved, {
			log: writeLog,
			missing: missing,
			stack: callback.stack
		}, null));
	}

	return onResolved(err, result);
});
```

test/missing.js
```diff
@@ -45,6 +51,46 @@ describe("missing", function() {
...
+		});
+		it("should report missing files exactly once when trying to resolve " + testCase[1], function(done) {
+			var callback = function(err, filename) {
+				if(err) {
+					var missing = err.missing.sort();
+					var isSame = true;
+					for(var i = 0; i < missing.length - 1; i += 2) {
+						isSame = isSame && (missing[i] === missing[i + 1]);
+					}
+					isSame.should.not.be.true('missing file names should not be repeated');
```
gdborton: Can you change this to use `"`?

gdborton:
@sokra Can you take another look?  This reduces our build times by roughly 30% from `2m47s` to `1m50s`, I added a `console.log` above `theLog.push` line, and it doesn't get called anymore.

@mikesherov Can you add the below to a .eslintrc file in `./test`? That'll allow for the testing globals that codacy is complaining about.

```json
{
  "globals": {
    "it": true,
    "describe": true,
    "beforeEach": true,
    "afterEach": true,
    "context": true,
    "before": true,
    "after": true,
  }
}
```

sokra:
Thanks

Mayank1791989:
@mikesherov @sokra  Thanks for the update. 
Tried new version today reduces my initial build time from ~60sec to ~28sec.

PR #92, opened 2017-06-27
introduce `cacheWithContext` option that excludes `request.context` from unsafeCache cache keys when false

mikesherov:


TheLarkInn:
Would this now be considered breaking?

mikesherov:
So, it's breaking on enhanced-resolve. What I can do in WP proper is detect if the user has provided any custom resolve or resolveLoader plugins, and if so, set the value of this new option to false.

mikesherov:
In that manner, we can do it in a semver patch way on Webpack proper.

mikesherov:
Ok yeah, I'm going to rename to `cacheWithContext` and default to `true` so it's semver minor.

mikesherov:
@TheLarkInn re-ready for re-review. It's now semver minor.

lukeapage:
I tried it out and got

```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration[0].resolve has an unknown property 'cacheWithContext'. These properties are valid:
   object { alias?, aliasFields?, cachePredicate?, descriptionFiles?, enforceExtension?, enforceModuleExtension?, extensions?, fileSystem?, mainFields?, mainFiles?, moduleExtensions?, modules?, plugins?, resolver?, symlinks?, unsafeCache?, useSyncFileSystemCalls? }
```

Do I need to also update a corresponding PR for the options validation?

lukeapage:
Ah found this PR: https://github.com/webpack/webpack/pull/5178/files

lukeapage:
webpack 3 - 65 seconds
webpack 3 with this PR - 34.4 seconds
webpack 2.3.3 - 34.0 seconds

:)

TheLarkInn:
I'm going to ship this. @sokra will have to publish as I don't yet have permissions on this repo to ship to npm.

PR #94, opened 2017-07-10
Add basic usage example and plugin example to README.md

DrewML:
Also made a few small other opinionated additions or modifications.


sokra:
Thanks

PR #96, opened 2017-07-20
Performance

sokra:


johnnyreilly:
Nice!  Is the performance gain from moving to using `Set` and `Map` instead of arrays and objects?

sokra:
yes. This PR is a sum of many small changes.

johnnyreilly:
Cool - something I'd like to move ts-loader to use Map - I didn't realize that could offer performance gains. Must take a look again

sokra:
Currently Map is slower than using object for mapping for all operation except delete (afaik). Here delete is used. It depends on the usecase.

sokra:
Also note that performance depends on node.js/v8 version. Older versions are slower with new ES6 stuff.

johnnyreilly:
Ah cool - I might hold off a couple months then

PR #103, opened 2017-07-24
Bug if key is not String

ldseinhardt:
Error with Laravel Mix in watch


zythum:
i got same problem. error at key.startWith when key is undefined. 
like: babel-loader check exists path, maybe is undefined.

sokra:
fixed by #100

PR #129, opened 2018-01-09
Check for multiple parts before accessing

lyndsey-ferguson:
This fixes an issue where the code throws an exception when it tries to access a `part` that does not exist. This occurs when `path` is `"."`.

I have used this in our application code and I have run `npm test` against these changes.

### DIFF
--- a/lib/getPaths.js
+++ b/lib/getPaths.js
@@ -9,16 +9,18 @@ module.exports = function getPaths(path) {
 	const paths = [path];
 	const seqments = [parts[parts.length - 1]];
 	let part = parts[parts.length - 1];
-	path = path.substr(0, path.length - part.length - 1);
-	paths.push(path);
-	for(let i = parts.length - 2; i > 2; i -= 2) {
-		part = parts[i];
-		path = path.substr(0, path.length - part.length) || "/";
+	if(parts.length > 1) {
+		path = path.substr(0, path.length - part.length - 1);
 		paths.push(path);
-		seqments.push(part.substr(0, part.length - 1));
+		for(let i = parts.length - 2; i > 2; i -= 2) {
+			part = parts[i];
+			path = path.substr(0, path.length - part.length) || "/";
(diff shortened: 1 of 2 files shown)


TheLarkInn:
Haven't formally reviewed this yet, but just for context was this a bottleneck you identified or a bug and if issue do you have an issue to tie to PR? Thanks for this comtrobution!!!!🙇🙇🙇🙇🙇

TheLarkInn:
Oh geeze I reread. Sorry I see you said there was an error. Could you please add tests guarding for this fix.

lyndsey-ferguson:
@TheLarkInn I've added tests to guard against the problem and the happy path.

lyndsey-ferguson:
@TheLarkInn is there anything else I need to do for this to move forward? Thanks.

lyndsey-ferguson:
@sokra can I get a review and a merge on this please?

sokra:
I wonder in which scenario `path` equals `.`. It's an absolute path.

lyndsey-ferguson:
@sokra we use `eslint-import-resolver-webpack` which, when testing a lint rule fails when this crashes. I admit that maybe `eslint-import-resolver-webpack` should not be sending relative paths, but I don't know how the internals work and if I fix this crash, it fixes our test. If you want a sample, I can put that together (I hadn't bothered as this seemed like a simple harmless fix).

lyndsey-ferguson:
@sokra is there anything else I need to do for this PR?

sokra:
I won't merge this PR, since this is an error in eslint-import-resolver-webpack.

enhanced-resolve should be passed an absolute path. Passing an relative path is wrong and will lead to weird side effects like `node_modules` can't be looked up correctly. (it's walking the directories upwards)

sokra:
But thanks for your work anyway.

lyndsey-ferguson:
Ok, we've patched this module so that it doesn't crash for us.

PR #132, opened 2018-01-31
Convert the codebase to ES6

ooflorent:


sokra:
Thanks

webmaster128:
Is node 6.11.5 a strict dependency? Or is 6.11.x sufficient?

ooflorent:
@webmaster128 It is the same as webpack. See https://github.com/webpack/webpack/blob/213226ece261a0e89edb9de15fbc6775418f6041/package.json#L71

montogeek:
According to http://node.green/ 6.4 would still be compatible (rest params, class, for..of)

PR #136, opened 2018-02-11
added resolveContext

Gcaufy:
When I use the demo code, I got an deprecated warning message.

https://github.com/webpack/enhanced-resolve/blob/master/lib/Resolver.js#L139


sokra:
Thanks

PR #135, opened 2018-02-09
sort options & added symlinks

Gcaufy:
1. Sort options follow the [document](https://webpack.js.org/configuration/resolve).
2. Added option `symlinks` from the [code](https://github.com/webpack/enhanced-resolve/blob/master/lib/ResolverFactory.js#L75).

PR #140, opened 2018-03-13
Fix symlinks default value on README

Deraen:
#135 added the symlinks option to readme but the correct default value is `true`, not `false`.


ooflorent:
Thanks!

Issue #141, opened 2018-03-26
Node engine version should be dropped to v.6.9.0


thomasmichaelwallace:
By specifying v.6.11.5 webpack excludes some typical environments, including the AWS Lambda one (currently at 6.10.3).

I looked back at the PR that caused this, and there doesn't seem to be an specific requirement. I've also prepared a PR that bumps it back down to 6.9 (specified by @TheLarkInn) and everything built and tested ok.

See this discussion for further details: https://github.com/webpack/webpack/issues/6579

ooflorent:
Closed by #142

PR #142, opened 2018-03-26
Drop node engine down to 6.9.0 to support AWS lambda environments.

thomasmichaelwallace:
Drops node engine version down, and confirms build/test against node 6.9.0.

This resolves issue #141, which in-turn will enable https://github.com/webpack/webpack/issues/6579 to be resolved.


ooflorent:
Thanks!

PR #145, opened 2018-03-27
Don't loop forever when faced with circular symlinks

jviotti:
We faced this issue when running Webpack on a Windows Concourse CI
(https://concourse-ci.org) worker.

In Windows, Concourse builds happen in directories that look like this:

```
C:\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes
```

On top of that, from within a build, Concourse symlinks C:\ to that same
volume directory that lives inside it. This means that when the
`SymlinkPlugin` kicks in, it will do something like this:

```
READLINK C:\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes -> undefined
READLINK C:\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b -> undefined
READLINK C:\concourse\work\volumes\live -> undefined
READLINK C:\concourse\work\volumes -> undefined
READLINK C:\concourse\work -> undefined
READLINK C:\concourse -> undefined
READLINK C: -> C:\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes
```

Therefore, the directory will be appended to itself over and over again,
leading to something like this:

```
C:\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes\concourse\work\volumes\live\e553aae5-ad96-4632-4970-959e0c11924b\volumes
```

The program will loop forever, until the host eventually runs out of
memory and crashes.

The fix is to include a set that tracks the visited directories, and
break the loop if we're going back to one of those.

Signed-off-by: Juan Cruz Viotti <jv@jviotti.com>

### DIFF
--- a/lib/SymlinkPlugin.js
+++ b/lib/SymlinkPlugin.js
@@ -22,15 +22,27 @@ module.exports = class SymlinkPlugin {
 			const paths = pathsResult.paths;
 
 			let containsSymlink = false;
+			const visitedPaths = new Set();
+
 			forEachBail.withIndex(paths, (path, idx, callback) => {
 				fs.readlink(path, (err, result) => {
 					if(!err && result) {
+						// Avoid circular links
+						if(visitedPaths.has(result)) {
+							return callback();
+						}
+
 						pathSeqments[idx] = result;
 						containsSymlink = true;
 						// Shortcut when absolute symlink found
 						if(/^(\/|[a-zA-z]:($|\\))/.test(result))
 							return callback(null, idx);
 					}
+
+					if(!err) {
+						visitedPaths.add(path);
+					}
+
 					callback();
 				});
 			}, (err, idx) => {
--- a/test/SymlinkPlugin.js
+++ b/test/SymlinkPlugin.js
@@ -1,0 +1,57 @@
+var fs = require("graceful-fs");
+var pathModule = require("path");
+var SymlinkPlugin = require("../lib/SymlinkPlugin");
+var ResolverFactory = require("../lib/ResolverFactory");
+
+describe("SymlinkPlugin: circular links", function() {
+	var resolver = ResolverFactory.createResolver({
+		extensions: [".js"],
+		fileSystem: {
+			stat: function(path, callback) {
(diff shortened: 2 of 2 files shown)


ooflorent:
What if we have the following:

```
$ ls /tmp/test
a -> /tmp/test/b
b -> /tmp/test/c
c
```

Does it prevent `/tmp/test/a` to be resolved to `/tmp/test/c`?

Instead of matching `^/` or `^[A-Z]:` you could use a `Set` to track visited locations. What do you think of this approach? Would it solve your issue?

jviotti:
Hi @ooflorent ,

Just to make sure I understand, the set of files and symbolic links from your example are:

```
/tmp/test/a -> /tmp/test/b
/tmp/test/b -> /tmp/test/c
/tpm/test/c -> undefined
```

Right? If so, given `/tmp/test/a`, the code will check if that's a symlink. Since that's the case, it should go to `/tmp/test/b` and check if that's a symlink. That's also the case, so then the code will go to `/tmp/test/c`, and start checking `/tmp/test`, `/tmp`, and `/`. None of those are symlinks in the proposed example, and therefore the result will correctly be `/tmp/test/c`. Does that sound right, or am I missing a key point in your example?

> Instead of matching ^/ or ^[A-Z]: you could use a Set to track visited locations. What do you think of this approach? Would it solve your issue?

That should work as well, and sounds like a more elegant solution. Let me update the PR to use this approach!

Thanks for the review!

jviotti:
@ooflorent I updated the PR to use the set based approach you recommended. Let me know what you think!

jviotti:
@ooflorent Is anything pending before we can merge this PR? :)

ooflorent:
@jviotti I would like another person to review the PR.

ooflorent:
I’d like @sokra or @TheLarkin to review it before.

sokra:
fixes by #150

PR #146, opened 2018-03-27
Code Refactoring

Manu-shukla:


ooflorent:
Thanks for your contribution but it doesn't match our coding style. We use multiple variable declaration statements inside of combining them.

Manu-shukla:
But, we can use this type of coding style, as it may be easier and much efficient.

PR #159, opened 2018-11-28
Update dev tools

ooflorent:
- Bump eslint
- Add prettier
- Add husky / lint-staged

Also uses config files from webpack


montogeek:
Looks good!

.prettierrc.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+	printWidth: 80,
```
montogeek: This it the default options

ooflorent: Yup. I did a copy of webpack's one.

PR #160, opened 2018-11-28
Remove compatibility layer for Tapable

ooflorent:
- `Resolver` no longer extends `Tapable`
- `Resolver#resolve` and `Resolver#doResolve` assume arguments are correct


sokra:
Thanks

PR #164, opened 2019-01-08
hoist readfile.bind

hulkish:
to support https://github.com/webpack/webpack/pull/8599


hulkish:
err nm, i can see this binding is already set

Issue #172, opened 2019-03-19
[Question] - custom fs implementation


dazinator:
Hello,

I would like to use an alternative underlying `fs` implementation, such as `memfs` or `unionfs`
I am not sure If I can do this, as [NodeJsInputFileSystem](https://github.com/webpack/enhanced-resolve/blob/95e27d33d052fe955d297c6165111cfd0cc03240/lib/NodeJsInputFileSystem.js) seems to be harcoded to use a particular underlying `fs` implementation. What would be the best way to go about this? Thanks

evilebottnawi:
Send a PR to allow setup own fs

dazinator:
Currently, the methods that access `fs` are defined at the prototype level: 

https://github.com/webpack/enhanced-resolve/blob/95e27d33d052fe955d297c6165111cfd0cc03240/lib/NodeJsInputFileSystem.js#L42

I think I'd need to remove that and have them declared at the class instance level, so that if a custom `fs` is passed in the constructor then that can be used, otherwise fall back to default `fs` implementation based on the `graceful-fs`. Does this sound ok with you?

evilebottnawi:
Yes

dazinator:
Great. PR Submitted.

sokra:
fixed, you can pass a `fileSystem` i. e. `fs` or `memfs`

PR #173, opened 2019-03-19
#172 - Support alternative fs implementations.

dazinator:
Allow an alternative fs implementation to be used, overriding the default based on `graceful-fs`

### DIFF
--- a/test/NodeJsInputFileSystem.js
+++ b/test/NodeJsInputFileSystem.js
@@ -3,11 +3,6 @@ var should = require("should");
 
 describe("NodeJsInputFileSystem", function() {
 	this.timeout(3000);
-	beforeEach(function() {});
-	afterEach(function() {
-		//fs.purge();
-	});
-
 	it("shoud not require fs argument in constructor", function(done) {
 		var nodeFs = new nodeJsInputFileSystem();
 		should.exist(nodeFs);
--- a/test/NodeJsInputFileSystem.js
+++ b/test/NodeJsInputFileSystem.js
@@ -1,0 +1,67 @@
+var nodeJsInputFileSystem = require("../lib/NodeJsInputFileSystem");
+var should = require("should");
+
+describe("NodeJsInputFileSystem", function() {
+	this.timeout(3000);
+	beforeEach(function() {});
+	afterEach(function() {
+		//fs.purge();
+	});
+
+	it("shoud not require fs argument in constructor", function(done) {
+		var nodeFs = new nodeJsInputFileSystem();
+		should.exist(nodeFs);
+		done();
+	});
+
+	it("shoud allow fs argument in constructor", function(done) {
+		const volumeDir = "app";
+		//const unionfs = require("unionfs");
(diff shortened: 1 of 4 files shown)


evilebottnawi:
Need add tests

dazinator:
I'm typically a dotnet developer, node environment is failry new to me. I can run tests via mocha etc. In terms of adding a new test, any guidance as to where I should add it? Do I create a new file for this issue under the tests directoy? Or should I append a new test to an existing file somewhere? Thanks

evilebottnawi:
Create new file (example of test https://github.com/webpack/enhanced-resolve/blob/master/test/CachedInputFileSystem.js), test existing method with and without own fs

dazinator:
I am seeing a lot of test failures on windows, i'll log under a seperate issue.

evilebottnawi:
@dazinator yep, we need fix their in separate issue, anyway feel free to send a PR with fixes too, a lot of developers work under linux so tests can be broken in some cases

dazinator:
Have added some tests.

dazinator:
I think this should meet the quality bar now but let me know if you need me to do anything further!

test/NodeJsInputFileSystem.js
```diff
@@ -0,0 +1,67 @@
+var nodeJsInputFileSystem = require("../lib/NodeJsInputFileSystem");
+var should = require("should");
+
+describe("NodeJsInputFileSystem", function() {
+	this.timeout(3000);
+	beforeEach(function() {});
+	afterEach(function() {
+		//fs.purge();
+	});
```
evilebottnawi: Need remove this

dazinator:
@evilebottnawi Have made change requested.

dazinator:
I did consider that. I guess it comes down to a choice.

If this class is changed to allow the 'fs' it works with to be injected, then it saves having to create essentially a duplicate.

I thought that creating a custom input file system could then be reserved for when:

1. you need to use something other than an 'fs' api to grab your files. Let's say you need to make a http call rec.
2. You actually have some custom logic to decide on the file 

I guess my thinking is that I wanted to save a consumer from having to create a custom input file system when neither of the above were true.

dazinator:
Any further action required from me here or is this awaiting a decision

sokra:
> Essentially widening the responsibility of the class from:
> 
> "This class is responsible for working with node's fs for file access'
> 
> to:
> 
> 'This class is responsible for working with an fs for file access"

The name is NodeJsFileSystem. It doesn't have a lot of logic. I don't think it makes sense to change this. sorry, but thanks for the PR anyway.

dazinator:
Ok thanks for considering. I'll create a duplicate class.

dazinator:
@sokra 
My only complaint is that having discussed this #172 and been told to submit a PR..
And then above being told to add tests..
Only then to be told the whole concept doesnt make sense to accept in the first place - sorry.
I know you must be busy but I wish you'd chimed in earlier!

PR #175, opened 2019-05-22
Add jail option

Globegitter:
This adds the `jail` option described in: https://github.com/webpack/webpack/issues/9165 Is there anything I need to change in webpack so it knows about this option? Or will it just pass it on as soon as a new `enhanced-resolve` is released?

Let me know if there is anything missing.

### DIFF
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -225,8 +225,12 @@ exports.createResolver = function(options) {
 			plugins.push(
 				new ModulesInHierachicDirectoriesPlugin("module", item, "resolve", jail)
 			);
-		else if (!jail || (jail && isSubDir(jail, item)))
+		else if (!jail || isSubDir(jail, item))
 			plugins.push(new ModulesInRootPlugin("module", item, "resolve"));
+		else
+			throw new Error(
+				`The provided module path '${item}' is not within the provided jail path '${jail}'. Please remove it.`
+			);
 	});
 
 	// relative
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -35,6 +35,9 @@ exports.createResolver = function(options) {
 	// A list of directories to resolve modules from, can be absolute path or folder name
 	let modules = options.modules || ["node_modules"];
 
+	// Limits module paths in hierarchic directories to be contained within the given path. Must be an absolute path.
+	const jail = options.jail;
+
 	// A list of description files to read from
 	const descriptionFiles = options.descriptionFiles || ["package.json"];
 
@@ -218,7 +221,7 @@ exports.createResolver = function(options) {
 	modules.forEach(item => {
 		if (Array.isArray(item))
 			plugins.push(
-				new ModulesInHierachicDirectoriesPlugin("module", item, "resolve")
+				new ModulesInHierachicDirectoriesPlugin("module", item, jail, "resolve")
 			);
 		else plugins.push(new ModulesInRootPlugin("module", item, "resolve"));
 	});
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -4,9 +4,9 @@
 */
 "use strict";
 
-const Resolver = require("./Resolver");
+const { sep } = require("path");
 
-const { isSubDir } = require("./getPaths");
+const Resolver = require("./Resolver");
 
 const SyncAsyncFileSystemDecorator = require("./SyncAsyncFileSystemDecorator");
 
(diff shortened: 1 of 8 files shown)


lib/getPaths.js
```diff
@@ -4,21 +4,39 @@
 */
 "use strict";
 
-module.exports = function getPaths(path) {
+const path = require("path");
+
+const isSubDir = function(parent, dir) {
+	const relative = path.relative(parent, dir);
```
sokra: path.relative is too slow. Can you implement this with `dir.startsWith(parent)` and add a `/` resp `\` to the jail option (in ResolverFactory)?

lib/ResolverFactory.js
```diff
@@ -218,9 +223,10 @@ exports.createResolver = function(options) {
 	modules.forEach(item => {
 		if (Array.isArray(item))
 			plugins.push(
-				new ModulesInHierachicDirectoriesPlugin("module", item, "resolve")
+				new ModulesInHierachicDirectoriesPlugin("module", item, "resolve", jail)
 			);
-		else plugins.push(new ModulesInRootPlugin("module", item, "resolve"));
+		else if (!jail || (jail && isSubDir(jail, item)))
+			plugins.push(new ModulesInRootPlugin("module", item, "resolve"));
```
sokra: if item is not part of jail we should consider this a configuration error and throw here.

lib/getPaths.js
```diff
@@ -4,21 +4,39 @@
...
+	return (
+		relative === "" ||
+		(relative && !relative.startsWith("..") && !path.isAbsolute(relative))
+	);
+};
+
+module.exports = function getPaths(path, jail) {
```
sokra: The SymlinkPlugin also uses getPaths and need to respect the jail option for symlinked values.

lib/getPaths.js
```diff
@@ -4,21 +4,39 @@
...
 	const parts = path.split(/(.*?[\\/]+)/);
-	const paths = [path];
-	const seqments = [parts[parts.length - 1]];
+	const paths = [];
+	const seqments = [];
 	let part = parts[parts.length - 1];
+	if (!jail || (jail && isSubDir(jail, path))) {
```
sokra: This can be an early return. Move it to the start of the function.

lib/getPaths.js
```diff
@@ -4,21 +4,39 @@
 */
 "use strict";
 
-module.exports = function getPaths(path) {
+const path = require("path");
+
+const isSubDir = function(parent, dir) {
```
sokra: This function doesn't fit into this file. But I guess we don't need it anymore when using `startsWith`.

lib/getPaths.js
```diff
@@ -4,21 +4,39 @@
...
 	let part = parts[parts.length - 1];
+	if (!jail || (jail && isSubDir(jail, path))) {
+		paths.push(path);
+		seqments.push(part);
+	}
 	path = path.substr(0, path.length - part.length - 1);
 	for (let i = parts.length - 2; i > 2; i -= 2) {
-		paths.push(path);
 		part = parts[i];
+		if (!jail || (jail && isSubDir(jail, path))) {
```
sokra: `isSubDir` can be avoided by comparing `path.length` with `jail.length`

sokra: You can `break` here, as path only gets shorter.

lib/getPaths.js
```diff
@@ -4,21 +4,39 @@
...
+			paths.push(path);
+			seqments.push(part.substr(0, part.length - 1));
+		}
 		path = path.substr(0, path.length - part.length) || "/";
-		seqments.push(part.substr(0, part.length - 1));
 	}
 	part = parts[1];
-	seqments.push(part);
-	paths.push(part);
+	if (!jail || (jail && isSubDir(jail, path))) {
```
sokra: length check instead

lib/ResolverFactory.js
Globegitter: The reason I had it like this is because I found that if one uses higher level projects like vue-cli or nuxt.js they set all sorts of paths in there and then this can just handled here transparently. But also happy to to fail as this can also be filtered out on the user side where we set the jail functionality.

sokra: yep. Better fail to prevent questions like "Why does it ignore the value passed to the modules option?".

lib/getPaths.js
Globegitter: Is that really needed if I just care about a jail for the module import? Doesn't the SymlinkPlugin also apply to non module paths? And how could I ensure the jail only applies to symlinked modules?

purkhusid:
Would love to see this get merged!

sokra:
fixed by #221

PR #181, opened 2019-06-26
feat: Accept multiple aliases

piglovesyou:
ref. https://github.com/webpack/webpack/issues/6817

Expect alias options to be:
```js
  // "Shorthanded" option type
  alias: {
    '@single@': './only',
    '@multi@': ['./first', './second'] // <- What this PR supports
  }
```

, which is equivalent to:

```js
  // "Raw" option type
  alias: [
    { name: '@single@', alias: './only' },
    { name: '@multi@', alias: './first' },
    { name: '@multi@', alias: './second' },
  ]
```

Later on I'll send another PR to [webpack/webpack](https://github.com/webpack/webpack) to support new alias option and to deal with https://github.com/webpack/webpack/issues/6817 .

### DIFF
--- a/lib/AliasPlugin.js
+++ b/lib/AliasPlugin.js
@@ -1,7 +1,7 @@
 /*
-	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
-*/
+ MIT License http://www.opensource.org/licenses/mit-license.php
+ Author Tobias Koppers @sokra
+ */
 "use strict";
 
 function startsWith(string, searchString) {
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Thanks for the PR
/cc @sokra

piglovesyou:
[I found this bug report](https://github.com/webpack/webpack/issues/5348). Should we simply support _raw type_ of the current option structure, instead of changing it? I would like to hear what you'd say, @sokra .

```js
  // "Raw" option type
  alias: [{
    name: '@single@',
    alias: './only',
  }, {
    name: '@multi@',
    alias: './first',
  }, {
    name: '@multi@',
    alias: './second',
  }]
```

piglovesyou:
Now I'm feeling https://github.com/webpack/enhanced-resolve/pull/181#issuecomment-506106863 is much better since it keeps API and simpler implementation. Sorry for my fussing, let me force push the another diff.

lfre:
@piglovesyou You beat me to it. We also have the need for this feature. I'd appreciate if this could also support the flat structure `[ path1, path2, path3 ]` you initially proposed. Here's the quick dirty way I did it: https://github.com/Hearst-Hatchery/enhanced-resolve/blob/master/lib/AliasPlugin.js#L45

piglovesyou:
@lfre Hello. Yeah initially I thought so but I ended up concluding it's enough just to support `[ path1, path2, path3 ]` only in _shorthanded option type_. Your code is nice but you restructure options in two places, don't you ([one in factory](https://github.com/Hearst-Hatchery/enhanced-resolve/blob/master/lib/ResolverFactory.js#L131) and [another in AliasPlugin](https://github.com/Hearst-Hatchery/enhanced-resolve/blob/master/lib/AliasPlugin.js#L46)). I assumed It's a bit tricky for maintainers.

If you have reasons to support `[ path1, path2, path3 ]` even in _raw option type_, I'd like to know.

lfre:
@piglovesyou I see what you mean, and I agree with your proposal here (shorthanded and raw). My intention was to change as little as possible, but it makes sense to simplify the `ResolverFactory` as well 👍. Looking forward to this.

piglovesyou:
@sokra , I'd like to hear what you think.

PR #185, opened 2019-07-06
gitignore .idea .vscode .DS_Store

vankop:
update gitignore


montogeek:
Why is this necessary?

vankop:
To prevent committing unnecessary files. I have made some PR's to webpack core repo and this folders are in .gitignore there.

Issue #184, opened 2019-07-06
Typescriptify enhanced-resolve


vankop:
## Feature request

It will be great to support typescript. Related to https://github.com/webpack/webpack/issues/3277 and https://github.com/webpack/webpack/issues/9196

**What is motivation or use case for adding/changing the behavior?**

documentation

**How should this be implemented in your opinion?**

Adding tsconfig, adding typescript as dev dependency. Maybe create typescript definitions as `webpack-source` does, something like `@types/webpack-enhanced-resolve`

**Are you willing to work on this yourself?**
yes

evilebottnawi:
PR welcome

sokra:
You can do it the same way as webpack uses types:

* JSDocs with type info
* lint step which runs `tsc --noEmit` to check the types

PR #192, opened 2019-07-23
esdoc resolve context

vankop:
https://github.com/webpack/enhanced-resolve/issues/184

What kind of change does this PR introduce?
esdoc `resolveContext`. Required for typization of resolver plugins.

Did you add tests for your changes?
N/A

Does this PR introduce a breaking change?
no

What needs to be documented once your changes are merged?
nothing

### DIFF
--- a/lib/Resolver.js
+++ b/lib/Resolver.js
@@ -13,11 +13,52 @@ const {
 	PathType
 } = require("./pathUtils");
 
+/** @typedef {import("tapable").Hook} Hook */
+/**
+ * @typedef {Object} ResolveRequest
+ * @property {string} path
+ * @property {string=} request
+ * @property {string=} query
+ * @property {boolean=} directory
+ * @property {boolean=} module
+ */
+/**
+ * String with special formatting
+ * @typedef {string} StackEntry
+ */
+/**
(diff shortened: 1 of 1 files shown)


lib/Resolver.js
```diff
@@ -13,11 +13,52 @@ const {
...
+ * @property {boolean=} module
+ */
+/**
+ * String with special formatting
+ * @typedef {string} StackEntry
+ */
+/**
+ * Resolve context
+ * @typedef {Object} ResolveContext
+ * @property {Set<string>} contextDependencies
```
vankop: did not find any adding to `contextDependencies`

lib/Resolver.js
```diff
@@ -13,11 +13,52 @@ const {
 	PathType
 } = require("./pathUtils");
 
+/** @typedef {import("tapable").Hook} Hook */
+/**
+ * @typedef {Object} ResolveRequest
```
vankop: It is not all possible request properties. I actually don't know is it required to define one big type since it could be some weird private fields see https://github.com/webpack/enhanced-resolve/blob/master/lib/getInnerRequest.js

PR #204, opened 2020-02-23
make trailing slash module path resolve correctly

Zuckjet:
Fixes [#8889](https://github.com/webpack/webpack/issues/8889)

### DIFF
--- a/lib/FileExistsPlugin.js
+++ b/lib/FileExistsPlugin.js
@@ -30,8 +30,6 @@ module.exports = class FileExistsPlugin {
 						if (resolveContext.log) resolveContext.log(file + " is not a file");
 						return callback();
 					}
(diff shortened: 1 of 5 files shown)


Zuckjet:
> Please add tests, thanks!

I dived into webpack source code , and found how to fix the bug, then I send this PR. 
But I had trouble in adding tests.  Indeed, I have tried several times, but seems none of them make sense.  should I add test in resolve.js like: 
```
testResolve(
		"file in module without extension",
		fixtures,
		"m4",
		path.join(fixtures, "node_modules", "m4", "index.js")
);
```
or add test for  plugin ?

@evilebottnawi  I would appreciate if you have any idea, thanks.

sokra:
Thanks

PR #209, opened 2020-03-10
Emit declarations

vankop:
add declarations

At first look types are ok =)


sokra:
fixed by new tooling, which started with this PR. Thanks

PR #211, opened 2020-03-28
docs(README): add table of contents

snitin315:
Enhancement,
Added table of contents in README.


sokra:
github has added that

PR #212, opened 2020-04-15
exports field test cases

vankop:


test/exportsField.js
```diff
@@ -0,0 +1,739 @@
...
+		}
+		//#endregion
+	];
+
+	testCases.forEach(testCase => {
+		it(testCase.name, () => {
+			if (testCase.expect instanceof Error) {
+				should.throws(() => processExportsField(...testCase.suite));
+			} else {
+				processExportsField(...testCase.suite).should.containDeepOrdered(
```
sokra, line 733: Why only contains and not eql?

test/exportsField.js
```diff
@@ -0,0 +1,739 @@
...
+			suite: [
+				{
+					".": {
+						browser: "./index.js",
+						node: "./src/node/index.js",
+						default: "./src/index.js"
+					}
+				},
+				".",
+				["default", "browser"]
```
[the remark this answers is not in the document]
vankop: What about allowing users to specify order themselves?

test/exportsField.js
```diff
@@ -0,0 +1,739 @@
...
+			]
+		},
+		{
+			name: "conditional mapping folder #3",
+			expect: ["lodash/index.mjs", "./utils/index.mjs", "./wpk/index.mjs"],
+			suite: [
+				{
+					"./utils/": {
+						webpack: "./wpk/",
+						browser: ["lodash", "./utils/"],
```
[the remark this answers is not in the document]
vankop: I followed spec and did not find answer.. intuitively you are correct since it is a folder mapping

test/exportsField.js
```diff
@@ -0,0 +1,739 @@
...
+						node: ["./node/"]
+					}
+				},
+				"./utils/index.mjs",
+				[]
+			]
+		},
+		{
+			name: "conditional mapping folder #3",
+			expect: ["lodash/index.mjs", "./utils/index.mjs", "./wpk/index.mjs"],
```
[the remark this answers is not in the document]
vankop: hmm.. maybe I did not get what you mean.. the order is `["browser", "webpack"]` so `browser` comes first. Keys ordering does not count

test/exportsField.js
vankop: Don't much familiar with `should.js`. fixed

vankop:
Small reviewing tests one more time, I think there are some things to add:
- from Node.js resolve algorithm `If target does not start with "./" or contains any "node_modules" segments including "node_modules" percent-encoding`, so probably `"node_modules" percent-encoding` needed
- incorrect backtracing, but positive resolve result, e.g.:
```
{
    expect: ["./dist/index"]
    suite: [
        {
              "./": "./src/../../",
              "./dist/": "./dist/"
         },
         "./dist/index",
         []
    ]
}
```

test/exportsField.js
```diff
@@ -0,0 +1,816 @@
...
+						node: "./node.js"
+					}
+				},
+				"./dist/index.js",
+				["browser"]
+			]
+		},
+		{
+			name: "Direct mapping #6",
+			expect: ["./index.js", "./src/index.js"],
```
sokra, line 159:
```suggestion
			expect: "./index.js",
```

test/exportsField.js
```diff
@@ -0,0 +1,816 @@
...
+						default: "./src/index.js"
+					}
+				},
+				".",
+				["browser"]
+			]
+		},
+		{
+			name: "Direct mapping #7",
+			expect: ["./src/index.js", "./index.js"],
```
sokra, line 174:
```suggestion
			expect: new Error(),
```

test/exportsField.js
```diff
@@ -0,0 +1,816 @@
...
+						browser: "/a/",
+						default: "/b/"
+					}
+				},
+				"./utils/index.mjs",
+				["browser"]
+			]
+		},
+		{
+			name: "incorrect exports field #6",
```
sokra, line 478:
```suggestion
			name: "incorrect request",
```

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+		);
+	}
+}
+
+/**
+ * @param {ConditionalMapping|Mapping} mapping mapping
+ * @returns {boolean} is conditional mapping
+ */
+function isConditionalMapping(mapping) {
+	return typeof mapping === "object" && !Array.isArray(mapping);
```
sokra, line 174: check for `null`

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+ * @param {string|string[]} directExport direct export
+ * @returns {string[]} mapping result
+ */
+function directMapping(remainingRequest, expectFolder, directExport) {
+	const isString = typeof directExport === "string";
+
+	if (isString) {
+		assertExport(/** @type {string} */ (directExport), expectFolder);
+	} else {
+		/** @type {string[]} */ (directExport).forEach(exp =>
```
sokra, line 189: `for(const exp of directExport)`

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	if (!expectFolder) {
+		return isString
+			? [/** @type {string} */ (directExport)]
+			: /** @type {string[]} */ (directExport).slice();
+	}
+
+	return isString
+		? [/** @type {string} */ (directExport.concat(remainingRequest))]
```
sokra, line 201:
```suggestion
		? [/** @type {string} */ (`${directExport}${remainingRequest}`)]
```

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/**
+ * @param {string} remainingRequest remaining request
+ * @param {boolean} expectFolder expect folder
+ * @param {string|string[]} directExport direct export
+ * @returns {string[]} mapping result
+ */
+function directMapping(remainingRequest, expectFolder, directExport) {
+	const isString = typeof directExport === "string";
```
sokra, line 184: Make a big `if (typeof directExport === "string") else` instead of storing it in a variable and testing it multiple times. This improves type inference and requires less `?:` here. (Less branches, better performance, but probably doesn't really matter here, is more for readability)

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+ * @param {ConditionalMapping|Mapping} mapping mapping
+ * @returns {boolean} is conditional mapping
+ */
+function isConditionalMapping(mapping) {
+	return typeof mapping === "object" && !Array.isArray(mapping);
+}
+
+/**
+ * @param {string} remainingRequest remaining request
+ * @param {boolean} expectFolder expect folder
```
sokra, line 179:
```suggestion
 * @param {string | undefined} remainingRequest remaining request when folder mapping, undefined for file mappings
```

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+	return isString
+		? [/** @type {string} */ (directExport.concat(remainingRequest))]
+		: /** @type {string[]} */ (directExport).map(d =>
+				d.concat(remainingRequest)
+		  );
+}
+
+/**
+ * @param {string} remainingRequest remaining request
+ * @param {boolean} expectFolder expect folder
```
sokra, line 209: Same as above

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Ivan Kopeykin @vankop
+*/
+
+"use strict";
+
+/** @typedef {string|string[]} Mapping */
+/** @typedef {{[k: string]: ConditionalMapping|Mapping}} ConditionalMapping */
```
sokra, line 9:
```suggestion
/** @typedef {Record<string, ConditionalMapping | Mapping>} ConditionalMapping */
```

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+/** @typedef {{[k: string]: ConditionalMapping|Mapping}|string} ExportField */
```
sokra, line 10:
```suggestion
/** @typedef {Record<string, ConditionalMapping | Mapping> | string} ExportField */
```

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+				// is nested
+				if (isConditionalMapping(mapping)) {
+					nextLookup = mapping;
+					continue loop;
+				}
+
+				return directMapping(
+					remainingRequest,
+					expectFolder,
+					/** @type {string|string[]} */ (mapping)
```
sokra, line 246:
```suggestion
					/** @type {Mapping} */ (mapping)
```

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+				);
+			}
+		}
+
+		nextLookup = null;
```
sokra, line 266:
``` json
{
  "exports": {
    "abc": { "def": "./x.js" },
    "ghi": "./y.js"
  }
}
```

With `abc` and `ghi` being active conditions, while `def` is not set.

from https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_nested_conditions

> If a nested conditional does not have any mapping it will continue checking the remaining conditions of the parent condition.

The result should be `["./y.js"]`.

Either use a lookup stack, or use recursion. I would go with recursion, as it's easier to implement here.

Another edge case test case:

``` json
{
  "exports": {
    "abc": { "def": "./x.js", "default": [] },
    "ghi": "./y.js"
  }
}
```

should result in `[]` with `abc` and `ghi`

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	return request.charCodeAt(i - 1) === slashCode ? i : -1;
+}
+
+/**
+ * @param {string} exp export target
+ * @param {boolean} expectFolder is folder expected
+ */
+function assertExport(exp, expectFolder) {
```
sokra, line 150: Would be cool if the error messages would include the value of `exp`.

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+	) {
+		throw new Error('Export should be relative path and start with "./"');
+	}
+
+	const isFolder = exp.charCodeAt(exp.length - 1) === slashCode;
+
+	if (isFolder !== expectFolder) {
+		throw new Error(
+			expectFolder
+				? "Expecting folder to folder mapping. Folder to file found"
```
sokra, line 163: Something like should end with `/` would be helpful

lib/processExportsField.js
```diff
@@ -0,0 +1,270 @@
...
+				: "Expecting file to file mapping. File to folder found"
```
sokra, line 164: Same, but must not end with `/`.

sokra:
Performance-wise this maybe needs a few optimization once everything is working. We can do this in the future in other PRs. The algorithmic complexity is a little bit high:

* For each resolved request -> `O(n)`
  * Sorting mappings in the exports object `O(m * log m)`
  * For each mapping `O(m)`
    * For each char in the key `O(p)`

=> `O(n * m * log m + n * m * p)` (n: number of requests to the package, m: number of mappings, p: avg length of requests)

n could be quite high, m could also be high for libraries like lodash

Due to caching in webpack n is limited to each different resolved request so it can be assumed to be `O(m)`.

p is too small to be relevant and bound by filesystem limitations

=> `O(m² * log m)`

So this could become a problem if someone uses a lot keys in the `exports` object.

I checked lodash, which has ~1000 files => m = 1000 => could be a problem.

---

A potential solution could be construct a [prefix tree (Trie)](https://en.wikipedia.org/wiki/Trie) for the exports object. This would make the sorting unnecessary and result could be retrieved in `O(p)`. Prefix tree only need to be constructed once per package.json (need to be cached) `O(q)` (q: number of packages)

=> `O(q * m * p + n * p)`

=> `O(q * m)`

basically only the cost of constructing the prefix trees is relevant as query cost is eliminated because n is limited to m with caching.

vankop:
> O(n * m * log m + n * m * p)

This is a worst case when all export field keys viewed

We can achieve `O(n + m * log m + n * m * p)` as a worst case if we will cache `cache.set(exportField, sortedExports)`

lib/processExportsField.js
vankop: typescript says Record can not circularly references itself

sokra:
> what about cost of construction?

I think it's `O(m * p)` but not sure.

jeffin143:
A trie , being implemented ❤️, I have only implemented or have seen it being used in competitive programming contests. 

Excited to this :)

Btw I guess 

> what about cost of construction?

it would `0(p*log(m))` if the trie is balanced BST , since the depth would be log(m) 

I am not sure

vankop:
> > what about cost of construction?
> 
> I think it's `O(m * p)` but not sure.

Yes, this is correct estimation (where p is average depth)

sokra:
Thanks. I merged it into a dev-1 branch where this feature could live until it's completed. You can do further PRs to this branch. We will merge it into master once it's finished.

One thing I noticed that is missing is the export sugar mentioned here: https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_conditional_exports
You can omit `"."` mapping and write a conditional mapping directly.
One can decide between these cases with: `all keys start with .` and `no key starts with .`. Any mix should be an error.

PR #213, opened 2020-05-03
handle syntax sugar for conditional mapping

vankop:
What kind of change does this PR introduce?
handle syntax sugar for conditional mapping

Did you add tests for your changes?
yes

Does this PR introduce a breaking change?
yes

What needs to be documented once your changes are merged?
None

### DIFF
--- a/lib/processExportsField.js
+++ b/lib/processExportsField.js
@@ -8,7 +8,7 @@
 /** @typedef {string|string[]} DirectMapping */
 /** @typedef {{[k: string]: MappingValue}} ConditionalMapping */
 /** @typedef {ConditionalMapping|DirectMapping|null} MappingValue */
-/** @typedef {Record<string, MappingValue>|string} ExportField */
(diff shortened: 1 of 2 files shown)


lib/processExportsField.js
```diff
@@ -8,7 +8,7 @@
 /** @typedef {string|string[]} DirectMapping */
 /** @typedef {{[k: string]: MappingValue}} ConditionalMapping */
 /** @typedef {ConditionalMapping|DirectMapping|null} MappingValue */
-/** @typedef {Record<string, MappingValue>|string} ExportField */
+/** @typedef {Record<string, MappingValue>|ConditionalMapping|string} ExportField */
```
sokra, line 11:
```suggestion
/** @typedef {Record<string, MappingValue>|ConditionalMapping|DirectMapping} ExportField */
```

I guess `"exports": ["./a", "./b"]` is also something valid that need to be handled.

sokra:
Thanks

PR #214, opened 2020-05-04
ExportFieldPlugin

vankop:


test/exportsField.js
```diff
@@ -923,3 +929,87 @@ describe("Exports field", function exportsField() {
...
+		resolver.resolve({}, fixture, "./lib/main.js", {}, (err, result) => {
+			if (err) return done(err);
+			if (!result) throw new Error("No result");
+			result.should.equal(path.resolve(fixture, "./lib/lib2/main.js"));
+			done();
+		});
+	});
+
+	it("throw error if extension not provided", done => {
+		resolver.resolve({}, fixture, "./lib/main", {}, (err, result) => {
```
sokra, line 968: As far as I understood the behavior these requests should not be affected by the exports field. Exports field should only applied for requests from outside: So only `exports-field/lib/main.js` but not `./lib/main.js` inside of `exports-field`.

vankop: Sure, I forgot about it..

vankop: > Exports field should only applied for requests from outside

There is also self-referencing, see test-cases

vankop:
Several problems here:

- to prevent `AliasFieldPlugin` use aliases I need to skip all `"file"` stage, but I need `FileExistsPlugin` should/can I add custom stage for this behavior?
- backtracking does not handled. Actually I don't think that this is a work of `ExportFieldPlugin`.. any ideas on this?
- self-referencing also a problem..

lib/Resolver.js
```diff
@@ -374,6 +376,7 @@ class Resolver {
 		if (part.request) {
 			part.module = this.isModule(part.request);
 			part.directory = this.isDirectory(part.request);
+			part.specifier = part.request[0] !== "." && part.request[0] !== "/";
```
sokra, line 379: No longer needed

test/fixtures/exports-field/node_modules/exports-field/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "exports-field",
+  "version": "1.0.0",
+  "main": "./main.js",
+  "browser": {
+    "./lib/main.js": "./lib/browser.js"
+  },
+  "exports": {
+    ".": "./index.js",
+    "./lib/": {
```
sokra, line 10:
```suggestion
    "./inner/": {
```

otherwise most test cases would also succeed if it doesn't work

lib/pathUtils.js
```diff
@@ -183,3 +183,32 @@ const cachedJoin = (rootPath, request) => {
 	return cacheEntry;
 };
 exports.cachedJoin = cachedJoin;
+
+const checkExportsFieldTarget = relativePath => {
```
vankop, line 187: I decided to write utility function for this for performance reasons. (checking "is request out of package scope" after normalization more expensive)

vankop:
Also as I understand
```
{
    "exports": {
          "node":  "./index.js",
          "default": "lodash/index.js"
    }
}
```
resolving `lodash/index.js` should work. Probably we need to pass appropriated hook to plugin to pipe such requests

sokra:
> Also as I understand
> 
> ```
> {
>     "exports": {
>           "node":  "./index.js",
>           "default": "lodash/index.js"
>     }
> }
> ```
> 
> resolving `lodash/index.js` should work. Probably we need to pass appropriated hook to plugin to pipe such requests

Is this behavior somewhere documented in the node.js docs or did you test that?

lib/SelfReferencePlugin.js
```diff
@@ -0,0 +1,74 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Ivan Kopeykin @vankop
```
vankop, line 3: this is incorrect =)

lib/SelfReferencePlugin.js
```diff
@@ -0,0 +1,74 @@
...
+				);
+				if (!exportsField) return callback();
+
+				const name = DescriptionFileUtils.getField(
+					request.descriptionFileData,
+					"name"
+				);
+				if (typeof name !== "string") return callback();
+
+				const req = request.request;
```
vankop, line 49: I think this can be checked before retrieving fields from description file

lib/SelfReferencePlugin.js
```diff
@@ -0,0 +1,74 @@
...
+				if (!req) return callback();
+
+				if (req === name || req.startsWith(`${name}/`)) {
```
vankop, line 52: maybe this will be a little bit better?
```suggestion
				if (req.startsWith(name)) {
				       if (req.length !== name.length && req[name.length] !== "/") return callback();
```

lib/ExportsFieldPlugin.js
```diff
@@ -38,35 +37,31 @@ module.exports = class ExportsFieldPlugin {
...
 				if (!exportsField) return callback();
 
-				const packageName = DescriptionFileUtils.getField(
-					request.descriptionFileData,
-					"name"
-				);
-				const requestStart = "./" + packageName;
-				const selfReferenced = initialRequest.startsWith(requestStart);
+				const remainingRequest = request.request;
+				if (remainingRequest === undefined) return callback();
```
vankop, line 56: probably this can be checked next to line 45

vankop:
Great that you found a lot of edge cases and bugs. I also thought about dividing logic of self referencing, but did not decide to challenge this and design with `initialRequest` probably not the best..

LGTM👍

sokra:
> > after that dev-1 can be merged into master, or what do you think?
> 
> Sure, maybe create an issue for improving README with `conditionNames`?

So it's not ready yet:

* [ ] document `conditionNames`
* [ ] add `exportsFields` option to configure exports fields
* [ ] document `exportsFields`

PR #219, opened 2020-05-13
Refactor array options to set

vankop:
I did refactoring also for options with `string|string[]`, I think this is useful too since mostly only strings will be used


evilebottnawi:
@vankop Big thanks!

lib/DescriptionFileUtils.js
```diff
@@ -56,7 +80,7 @@ function loadDescriptionFile(
 						}
 						let json;
 						try {
-							json = JSON.parse(content);
+							json = JSON.parse(/** @type {string} */ (content));
```
sokra, line 83: It could be a Buffer too, so the cast is not correct. Call .toString() that should work for Buffer and string

lib/Resolver.js
```diff
@@ -22,11 +22,20 @@ const {
  * @property {function(): boolean} isFile
  */
 
+/**
+ * @typedef {Object} FileSystemError
```
sokra, line 26: Maybe call it `MaybeFileSystemError` since all properties are optional.

lib/ResolverFactory.js
```diff
@@ -66,19 +66,19 @@ const UseFilePlugin = require("./UseFilePlugin");
...
+ * @property {Set<string>} descriptionFiles
  * @property {boolean} enforceExtension
- * @property {string[]} extensions
+ * @property {Set<string>} extensions
  * @property {FileSystem} fileSystem
  * @property {Object | false} unsafeCache
  * @property {boolean} symlinks
  * @property {Resolver=} resolver
- * @property {(string | string[])[]} modules
+ * @property {Set<string | string[]>} modules
```
sokra, line 79: Arrays in Sets are a little bit weird, but since this is only for performance I think we can ignore that and accept duplication here. Arrays are very rare anyway.

lib/ResolverFactory.js
```diff
@@ -124,11 +144,13 @@ function createOptions(options) {
...
+		descriptionFiles: new Set(options.descriptionFiles || ["package.json"]),
 		enforceExtension: options.enforceExtension || false,
-		extensions: options.extensions
-			? /** @type {string[]} */ ([]).concat(options.extensions)
-			: [".js", ".json", ".node"],
+		extensions: new Set(
+			options.extensions
+				? /** @type {string[]} */ ([]).concat(options.extensions)
+				: [".js", ".json", ".node"]
+		),
```
sokra, line 153: Can be simplified to `options.extensions || [".js", ".json", ".node"]`

lib/ResolverFactory.js
```diff
@@ -434,21 +439,23 @@ function aliasOptionsToArray(alias) {
...
  */
-function mergeFilteredToArray(array, filter) {
-	return array.reduce((/** @type {Array<string | string[]>} */ array, item) => {
-		if (filter(item)) {
-			const lastElement = /** @type {string[]} */ (array[array.length - 1]);
-			if (Array.isArray(lastElement)) {
-				lastElement.push(item);
+function mergeFilteredToSet(array, filter) {
+	return new Set(
+		array.reduce((/** @type {Array<string | string[]>} */ array, item) => {
```
sokra, line 446: Set need to be applied to the incoming `array`. Due to the nested structure a Set as result is not needed/has not the expected effect.

lib/ResolverFactory.js
vankop: yep, I mentioned this in PR description

lib/ResolverFactory.js
```diff
@@ -124,11 +144,11 @@ function createOptions(options) {
 			typeof options.cacheWithContext !== "undefined"
 				? options.cacheWithContext
 				: true,
-		descriptionFiles: options.descriptionFiles || ["package.json"],
+		descriptionFiles: Array.from(
```
vankop, line 147: we can do `Array.from` later on line 206 to keep here consistent Set options, but I decided to keep optimization logic in normalization.

PR #221, opened 2020-05-31
add restrictions option support

vankop:


test/restrictions.js
```diff
@@ -0,0 +1,38 @@
...
+		restrictions: [fixture]
+	});
+
+	resolver.resolve({}, fixture, "pck2", {}, (err, result) => {
+		if (!err) throw new Error(`expect error, got ${result}`);
+		err.should.be.instanceof(Error);
+		err.message.should.match(/Resolve restriction ".*" not passed/);
+		done();
+	});
+});
```
evilebottnawi, line 38: Can we add tests (with the `restrictions` option) where:
- we have `.css` in `extensions` - should resolve `index.css`
- we have `style` in `mainFields` - should resolve `index.css`
- we have `index.css` - `package/index.css`, `package/index.js`, `package.json` with `main: 'index.js'` (`mainFiles`) - should resolve `index.css`

lib/ResolverFactory.js
```diff
@@ -241,6 +246,7 @@ exports.createResolver = function(options) {
 	resolver.ensureHook("file");
 	resolver.ensureHook("finalFile");
 	resolver.ensureHook("existingFile");
+	resolver.ensureHook("finalResolved");
```
sokra, line 249: I think you don't need a new hook here. You could place the restrictions plugin before the result plugin in the resolved hook.

In this case the restrictions plugin don't need a `target`, it would only error when needed and pass the control flow further (with `callback()`) when everything is fine.

test/restrictions.js
sokra: I think to get the expected behavior (resolving should try alternatives when hitting a restriction) the restrictions plugin need to return with `null` instead of throwing an error. An error would cancel the full resolving.

You can "log" (resolveContext.log) an message that a restriction was hit.

vankop: Ah, I misunderstand expected behavior, fixed

sokra:
Looks good. Need to review in detail when on computer...

sokra:
Thanks

PR #225, opened 2020-06-09
support query in ExportsFieldPlugin

vankop:
- add exports field test cases
- add types export

### DIFF
--- a/lib/ExportsFieldPlugin.js
+++ b/lib/ExportsFieldPlugin.js
@@ -43,15 +43,17 @@ module.exports = class ExportsFieldPlugin {
 			.tapAsync("ExportsFieldPlugin", (request, resolveContext, callback) => {
 				// When there is no description file, abort
 				if (!request.descriptionFilePath) return callback();
-				// When the description file is inherited from parent, abort
-				// (There is no description file inside of this package)
-				if (request.relativePath !== ".") return callback();
+				if (
+					// When the description file is inherited from parent, abort
+					// (There is no description file inside of this package)
+					request.relativePath !== "." ||
+					request.request === undefined
+				)
+					return callback();
 
 				const remainingRequest = request.query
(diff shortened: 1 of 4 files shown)


lib/ExportsFieldPlugin.js
```diff
@@ -47,7 +47,9 @@ module.exports = class ExportsFieldPlugin {
 				// (There is no description file inside of this package)
 				if (request.relativePath !== ".") return callback();
 
-				const remainingRequest = request.request;
+				const remainingRequest = request.query
+					? (request.request === "." ? "./" : request.request) + request.query
+					: request.request;
 				if (remainingRequest === undefined) return callback();
```
sokra, line 53: You should move this line one up to keep the expected behavior

lib/ExportsFieldPlugin.js
```diff
@@ -102,7 +104,17 @@ module.exports = class ExportsFieldPlugin {
 				forEachBail(
 					paths,
 					(p, callback) => {
-						const error = checkExportsFieldTarget(p);
+						const queryIndex = p.lastIndexOf("?");
```
sokra, line 107:
```suggestion
						const queryIndex = p.indexOf("?");
```

query string could contain `?` itself

test/exportsField.js
```diff
@@ -130,6 +130,183 @@ describe("Process exports field", function exportsField() {
...
+					"./🎉": "./ok.js",
+					"./%F0%9F%8E%89": "./other.js",
+					"./bar#foo": "./ok.js",
+					"./#zapp/": "./"
+				},
+				"./module?foo",
+				[]
+			]
+		},
 		//#endregion
```
sokra, line 310: add a test case for `{ "./a?b?c/": "./" }` `"./a?b?c/d?e?f"` -> `./d?e?f`

Looks weird I know^^

sokra:
Thanks

PR #227, opened 2020-07-02
add roots option

vankop:


types.d.ts
```diff
@@ -392,6 +398,7 @@ declare namespace exports {
 		PnpApiImpl as PnpApi,
 		Resolver,
 		FileSystem,
+		ResolveContext,
```
vankop, line 401: this was removed in original definitions in webpack
https://github.com/webpack/webpack/commit/8174145782c1e996f03e3d94d3b913502958e2a3#diff-c3dfe281c86f25f2e019a00e2945d598L360
So I added export here to keep types consistent

sokra:
Thanks

PR #232, opened 2020-07-16
allow to flag requests as already fully specified

sokra:
this is needed for compatibility with how node.js resolves requests in ESM


vankop:
interesting, why `{extensions: []}` not suitable?

sokra:
I tried extensions: [], but there is an edge case in the node.js impl regarding resolving packages. For the main field extensions and directories are resolved the legacy way.

PR #237, opened 2020-07-26
add imports field support

vankop:


lib/util/entrypoints.js
```diff
@@ -0,0 +1,605 @@
...
+	while (slashIndex !== -1) {
+		const folder = request.slice(lastNonSlashIndex, slashIndex);
+
+		const folderMapping = node.folder;
+		if (folderMapping) {
+			if (lastFolderMatch) {
+				lastFolderMatch[0] = folderMapping;
+				lastFolderMatch[1] = lastNonSlashIndex;
+			} else {
+				lastFolderMatch = [folderMapping, lastNonSlashIndex || 2];
```
vankop, line 278: this is needed to concatenate without "./". This matching could be implemented other way with request "normalization" (without "./"), but this will require slicing request string and some extra logic too. So I think this design is fine enough and performs faster (less allocations)

lib/util/entrypoints.js
```diff
@@ -0,0 +1,605 @@
...
+}
+
+/**
+ * Internal helper for building path tree
+ * @param {PathTreeNode} root root
+ * @param {string} path path
+ * @param {MappingValue} target target
+ */
+function walkPath(root, path, target) {
+	if (path.length === 2 && path === "./") {
```
vankop, line 447: this is needed to prevent handling "./" as file..

lib/ResolverFactory.js
```diff
@@ -311,11 +317,33 @@ exports.createResolver = function(options) {
...
 
+	// internal
+	importsFields.forEach(importsField => {
+		plugins.push(
+			new ImportsFieldPlugin(
+				"internal",
+				conditionNames,
+				importsField,
+				"final-file",
+				"internal-resolve"
```
vankop, line 342: technically we can use `raw-module` here

vankop:
Forgot to change hash parsing, will fix it..

sokra (approved):
Not fully reviewed, but I trust you that this is fine :)

PR #241, opened 2020-08-27
add fallback option

vankop:
- rework how AliasPlugin work
- add fallback option and test cases


sokra:
Thanks

vankop:
yeh, that was easy..

PR #252, opened 2020-09-28
avoid stack overflow due to cached resolving

sokra:
Fixes https://github.com/webpack/webpack/issues/11518


vankop:
will be cool to find a test case..

vankop (approved):

PR #259, opened 2020-10-14
Update getType in path.js to handle null paths

gtwilliams03:
Fix per discussion in webpack #11668


evilebottnawi (changes requested):
Invalid fix, we should fix the problem when `p` is null/undefined, instead suppress it, because it will not work

alexander-akait:
Closing due to inactivity. Please test with latest version and feel free to reopen if still regressions. Thanks!

PR #262, opened 2020-11-03
chore(deps): actually use graceful-fs

AviVahl:
Usage was removed at https://github.com/webpack/enhanced-resolve/pull/182

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,7 +5,7 @@
 
 "use strict";
 
-const fs = require("fs");
+const fs = require("graceful-fs");
 const CachedInputFileSystem = require("./CachedInputFileSystem");
 const ResolverFactory = require("./ResolverFactory");
 
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "LICENSE"
   ],
   "dependencies": {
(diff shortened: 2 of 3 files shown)


evilebottnawi (approved):
Good catch

sokra (changes requested):
Good catch.

I think it was a mistake that this was removed.

This line

https://github.com/webpack/enhanced-resolve/blob/d6807b553dc7e3157746b57a82cb238cd5e58cee/lib/index.js#L8

should require graceful-fs instead.

Do you want to change that?

evilebottnawi (approved):

AviVahl:
I changed it back per @sokra's comment.

TBH, as a *standalone* library, I would have expected enhanced-resolve to not use/require `graceful-fs` by default...
When used in `webpack`, it could have provided it with its own copy of `graceful-fs`... (via the `fileSystem` option)

evilebottnawi:
Yep, we can improve it, but let's it do in the separate PR

sokra (approved):

AviVahl:
@sokra this is ready for a merge, I guess...?

sokra:
yes

sokra:
thanks

PR #266, opened 2020-12-02
remove fullySpecified when resolving a module request with PnP

sokra:
fixes https://github.com/webpack/webpack/issues/12093

### DIFF
--- a/lib/PnpPlugin.js
+++ b/lib/PnpPlugin.js
@@ -77,7 +77,9 @@ module.exports = class PnpPlugin {
 					...request,
 					path: resolution,
 					request: undefined,
-					ignoreSymlinks: true
+					ignoreSymlinks: true,
+					fullySpecified:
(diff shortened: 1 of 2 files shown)


johnrichardrinehart:
What does

> request.fullySpecified && !/^(@[^/]+\/)?[^/]+$/.test(req)

mean, in English? That regex is a mouthful.

sokra:
It tests for a package e.g `package` or `@user/package` without subpath within the package like `package/path`

johnrichardrinehart:
Oh, yep. That's what it does. The capture group begins at the start of the string, grabs `@me/a/b/c` in a non-greedy way so that you don't capture `@me/a/b/c`, you only match `@me/a`. Then you look for any number of non-slash characters. So, `@me/a/b/c` would fail because `/b/c` exists.

Got it. Thanks!

PR #267, opened 2020-12-04
Minimize footprint for browser bundling

barak007:
Allow `enhanced-resolve/lib/ResolverFactory.js` to be bundled for the browser.

After this change the the only requirement for browser bundling is node's `path` package which is reasonable.

### DIFF
--- a/lib/CachedInputFileSystem.js
+++ b/lib/CachedInputFileSystem.js
@@ -8,6 +8,8 @@
 /** @typedef {import("./Resolver").FileSystem} FileSystem */
 /** @typedef {import("./Resolver").SyncFileSystem} SyncFileSystem */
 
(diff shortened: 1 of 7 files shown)


barak007:
It seems like travis-ci is stuck.

barak007:
@sokra this is a continuation of the PR  in tapable

alexander-akait:
Something wrong with types :confused:

sokra:
Thanks

alexander-akait:
@barak007 Just interesting what is use case? bundle webpack for browser?

barak007:
I did bundle webpack for browser once and it worked fine... This time it's related to our Stylable library that uses enhanced-resolve and I need it to work in browser.

Issue #263, opened 2020-11-14
Can't resolve 'pnpapi' with webpack 5.4.0


dstampher:
```
ERROR in ./node_modules/enhanced-resolve/lib/ResolverFactory.js 118:9-26
Module not found: Error: Can't resolve 'pnpapi' in 'C:\Users\david\Projects\electron-react @ ./node_modules/enhanced-resolve/lib/index.js 10:24-52
 @ ./node_modules/webpack/lib/ResolverFactory.js 8:16-59
 @ ./node_modules/webpack/lib/Compiler.js 25:24-52
 @ ./node_modules/webpack/lib/index.js 114:9-30
 @ ./src/js/App.js 2:0-35
 @ ./src/js/index.js 3:0-24 5:50-53
```

Here's my package.json as well

```
{
  "name": "electron-react",
  "version": "1.0.0",
  "description": "",
  "main": "main.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "watch": "webpack --config webpack.common.js --watch",
    "start": "electron ."
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "electron": "^10.1.5",
    "react": "^17.0.1",
    "react-dom": "^17.0.1"
  },
  "devDependencies": {
    "@babel/core": "^7.12.3",
    "@babel/preset-env": "^7.12.1",
    "@babel/preset-react": "^7.12.5",
    "babel-loader": "^8.2.1",
    "css-loader": "^5.0.1",
    "electron-reload": "^1.5.0",
    "sass": "^1.29.0",
    "sass-loader": "^10.1.0",
    "style-loader": "^2.0.0",
    "webpack": "^5.4.0",
    "webpack-cli": "^4.2.0"
  }
}
```

From what I've heard, pnpapi is built into webpack now, and that might be causing some conflict?  I commented out line 118 in lib/ResolverFactory.js and webpack compiles fine after that.

sokra:
Whats in this line? src/js/App.js 2:0-35

dstampher:
@sokra ResolverFactory.js 118

BradBailey-web:
I stopped getting the aformentioned error by changing 

`./node_modules/webpack/node_modules/enhanced-resolve/lib/ResolverFactory.js 118:9-26`
to 
`return require("./PnpPlugin").PnpApiImpl`

RAMKUMARDANE:
[...]

PR #286, opened 2021-02-21
Update ResolverFactory.js

dlredden:
Issue #263 is preventing my company from upgrading to Webpack v5. This was the proposed fix that worked well for us so I created a PR.

### DIFF
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -118,7 +118,7 @@ function processPnpApiOption(option) {
 		/** @type {NodeJS.ProcessVersions & {pnp: string}} */ versions.pnp
 	) {
 		// @ts-ignore
-		return require("pnpapi"); // eslint-disable-line node/no-missing-require
+		return require("./PnpPlugin").PnpApiImpl; // eslint-disable-line node/no-missing-require
 	}
 
 	return option || null;


alexander-akait:
Please provide example how we can reproduce the problem

ferdinando-ferreira:
> Please provide example how we can reproduce the problem

@alexander-akait : https://github.com/ferdinando-ferreira/enhanced-resolve-pnpapi-bug

```
git clone https://github.com/ferdinando-ferreira/enhanced-resolve-pnpapi-bug.git enhanced-resolve-pnpapi-bug
cd enhanced-resolve-pnpapi-bug
npm install
npm run build
```

alexander-akait:
@ferdinando-ferreira Why do you bundle webpack? You should ignore this module (`resolve.fallback.pnpapi: false`) when you bundle webpack for browser, because browsers doesn't support pnp

ferdinando-ferreira:
> Why do you bundle webpack?

@alexander-akait : To provide the minimal usecase requested. 

Regardless of usefulness it provides the exact error case reported:

`require("pnpapi")`

when it should be

`require("./PnpPlugin").PnpApiImpl`

because there is no pnpapi package.

Would you prefer a "realistic" use case that would hit the same error?

alexander-akait:
Please read https://yarnpkg.com/advanced/pnpapi, "pnpapi" !== "require("./PnpPlugin").PnpApiImpl"

ferdinando-ferreira:
> Please read https://yarnpkg.com/advanced/pnpapi, `require("pnpapi")` !== `require("./PnpPlugin").PnpApiImpl`

You are correct! The root cause of the error must be then the original issue submitter accidentally bundling something with a require to webpack (or other package depending on `enhanced-resolve`).

@dlredden, @dstampher: any chance that's the case?

alexander-akait:
Yep, in this case `resolve.fallback.pnpapi: false` should solve the problem

sokra:
It's not `resolve.fallback` but `externals` instead:

``` js
externals: {
  pnpapi: true
}
```

Issue #273, opened 2020-12-22
Bug: incorrectly resolves using node_modules when using PnP


merceyz:
**Describe the bug**

When running under Yarn PnP `enhanced-resolve` will incorrectly use the `node_modules` resolution

**To Reproduce**

Setup:
```sh
mkdir repro && cd repro
yarn init -y
yarn set version berry
yarn add enhanced-resolve@5.4.1 lodash@4.17.20
```
Code:

```ts
import { ResolverFactory, CachedInputFileSystem } from 'enhanced-resolve';
import path from 'path';
import fs from 'fs';

const cachedFS = new CachedInputFileSystem(fs);

const commonResolver = ResolverFactory.createResolver({
  fileSystem: cachedFS,
  useSyncFileSystemCalls: true,
});

fs.mkdirSync(path.join(__dirname, 'node_modules/lodash'), { recursive: true });
fs.writeFileSync(path.join(__dirname, 'node_modules/lodash/foo.js'), '');

commonResolver.resolveSync({}, __filename, 'lodash/foo');
throw new Error(`Resolve should have failed since lodash/foo doesn't exist`);

```

alexander-akait:
Expected, because `modules: ['node_modules']` by default, set `modules: []`

merceyz:
I originally did that, but that causes it to throw on resolving `lodash`

alexander-akait:
What is the real case problem?

merceyz:
I'm adding `exports` support to the PnP api so I need `enhanced-resolve` to only do the resolution from the result of `resolveToUnqualified`, when `resolveToUnqualified` returns a path, if the file `enhanced-resolve` is looking for doesn't exist it will keep on climbing the filetree and potentially return a path from anything else

alexander-akait:
Please describe the problem, not code, `exports` in `package.json`?

merceyz:
The problem is that `enhanced-resolve` walks up the filetree and out of the `node_modules` folder given to it by PnP

> exports in package.json?

Yes, the pnp patch doesn't support it at the moment, it only respects the `main` field

alexander-akait:
> The problem is that enhanced-resolve walks up the filetree and out of the node_modules folder given to it by PnP

So you expected throw error?

> Yes, the pnp patch doesn't support it at the moment, it only respects the main field

It was fixed in the latest version, webpack supports `exports` for PnP

merceyz:
> So you expected throw error?

[...]

PR #289, opened 2021-04-15
fix: give precedence to pnp over node_modules, close #288

### LINKED ISSUE #288: Bug: linked package incorrectly uses node_modules instead of pnpapi
enhanced-resolve is incorrectly escaping from the pnpapi context.

### Context

Encountered this via webpack initially using Yarn 2's pnpapi.

### With Webpack:

Repro project: https://github.com/noahnu/yarn-repro-linking-non-pnp, after cloning the project, `cd yarn2-pnp-project && yarn build` then search for `yarn1-nm-project/node_modules/react-dom/index.js` in the `dist/bundle.js` file. It's from node_modules.

And without webpack, purely using enhanced-resolve: 

### Repro:

Clone: https://github.com/noahnu/yarn-repro-linking-non-pnp/tree/pure_enhance

and run the bash script: `./run-repro.sh`.

```js
const assert = require('assert')
const resolve = require('enhanced-resolve')

assert.equal(
  resolve.sync(require.resolve('my-linked-package'), 'react-dom'),
  require.resolve('react-dom', {
    paths: [require.resolve('my-linked-package')],
  })
)
```

### Expectation

I expect the assertion to pass, as both should be going through pnp.

noahnu:
Fixes https://github.com/webpack/enhanced-resolve/issues/288.

Ideally, node_modules should not even be considered if the pnp api is enabled, however it looks like that's intentional according to https://github.com/webpack/enhanced-resolve/issues/273

### DIFF
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -399,14 +399,14 @@ exports.createResolver = function (options) {
 	});
 	modules.forEach(item => {
 		if (Array.isArray(item)) {
-			plugins.push(
-				new ModulesInHierachicDirectoriesPlugin("raw-module", item, "module")
-			);
 			if (item.includes("node_modules") && pnpApi) {
 				plugins.push(
 					new PnpPlugin("raw-module", pnpApi, "undescribed-resolve-in-package")
 				);
 			}
+			plugins.push(
+				new ModulesInHierachicDirectoriesPlugin("raw-module", item, "module")
+			);
 		} else {
 			plugins.push(new ModulesInRootPlugin("raw-module", item, "module"));
 		}
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -401,12 +401,20 @@ exports.createResolver = function (options) {
 		if (Array.isArray(item)) {
 			if (item.includes("node_modules") && pnpApi) {
 				plugins.push(
+					new ModulesInHierachicDirectoriesPlugin(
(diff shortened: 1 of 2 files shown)


alexander-akait:
hm, why developers have `node_modules` and use `yarn PnP`, I think yarn should print error/warning here

merceyz:
It happens when using the `portal:` protocol to a project that uses `node_modules`, for example Next.js uses Yarn 1 with `node_modules` but has e2e tests using Yarn 2 with PnP that uses the built sources from the Yarn 1 project

https://github.com/vercel/next.js/blob/15133b47ec7bfc413aa068d92fb678626cf47571/test-pnp.sh#L40

noahnu:
Another example which prompted this PR/issue:

At my work we have 2 projects. Project A using yarn pnp and Project M using node_modules. Project M is a monorepo. Right now if you try to link a package from Project M into Project A, the peer dependencies of the Project M package are resolved using Project M's node_modules, rather than Project A's dependencies. If the peer deps include React, you get singleton issues preventing React hooks from working, among other features.

merceyz:
Since Next.js started using webpack 5 by default this is also affecting their e2e tests https://github.com/vercel/next.js/runs/2373726821#step:5:222, applying the change in this PR fixes their e2e tests cc @sokra

sokra (approved):

sokra:
Thanks

PR #312, opened 2021-10-01
fix: fix the wrong path when request dirname is same as root

### LINKED ISSUE #311: RootsPlugin causes watchpack watches a set of wrong paths
https://github.com/webpack/enhanced-resolve/blob/d1a3cb65cdf90754871a263172006e2902a2f86e/lib/RootsPlugin.js#L42

This gives a path `'/Users/wood/Desktop/webpack-bug/Users/wood/Desktop/webpack-bug/examples/relay/src/index.web'` See the following screenshot

image

because of that, watchpack watches a set of wrong paths

https://github.com/webpack/watchpack/blob/c651129195d84cc80e547a45be38861b5bbada1a/lib/watchEventSource.js#L56

image

Is there a reason of coding this `resolver.join(root, req.slice(1));`? 

Here is the reproducible repo https://github.com/wood1986/rootplugin-bug
1. yarn
2. yarn start

wood1986:
fix #311

### DIFF
--- a/lib/RootsPlugin.js
+++ b/lib/RootsPlugin.js
@@ -39,7 +39,10 @@ class RootsPlugin {
 				forEachBail(
 					this.roots,
 					(root, callback) => {
-						const path = resolver.join(root, req.slice(1));
+						const path = req.startsWith(root)
+							? req
+							: resolver.join(root, req.slice(1));
+
 						const obj = {
 							...request,
 							path,
--- a/test/roots.js
+++ b/test/roots.js
@@ -108,6 +108,31 @@ describe("roots", () => {
 		);
 	});
 
+	it("should support roots options with resolveToContext and absolute request path", done => {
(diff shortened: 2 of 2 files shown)


lib/RootsPlugin.js
```diff
@@ -39,7 +39,10 @@ class RootsPlugin {
 				forEachBail(
 					this.roots,
 					(root, callback) => {
-						const path = resolver.join(root, req.slice(1));
+						const path = root.includes(req)
```
alexander-akait, line 42: What about `/foo/bar/foo/bar`?

wood1986: it should be startsWith

wood1986:
@alexander-akait Could you please review again?

alexander-akait (commented):
Again, you can have `/foo/bar/foo/bar`, it is possible, fix is invalid, we should check `/foo/bar/` and `/foo/bar/foo/bar`

PR #320, opened 2021-12-10
chore: add cSpell for spell checking

nschonni:
Add cSpell and fix minor typos


.cspell.json
```diff
@@ -0,0 +1,38 @@
...
+    "addrs",
+    "anotherhashishere",
+    "arcanis",
+    "Builtins",
+    "complexm",
+    "endregion",
+    "entrypoints",
+    "filesystems",
+    "hashishere",
+    "Hierachic",
```
nschonni, line 14: This is actually a typo in ModulesInHierachicDirectoriesPlugin, but will be a breaking change to fix

lib/ResolverFactory.js
```diff
@@ -280,7 +280,7 @@ exports.createResolver = function (options) {
 
 	resolver.ensureHook("resolve");
 	resolver.ensureHook("internalResolve");
-	resolver.ensureHook("newInteralResolve");
+	resolver.ensureHook("newInternalResolve");
```
alexander-akait, line 283: Potential breaking change, better to add deprecation

nschonni: How would a depreciation work for this one?
PS: did a quick search and it doesn't look like it's referenced anywhere else https://github.com/search?q=org%3Awebpack+Interal&type=code

alexander-akait: This package used not only in webpack, many tools use it for resolving

sokra:
Thanks

PR #323, opened 2021-12-10
chore: upgrade to Husky 6

nschonni:


nschonni:
Downgraded to 7 like the core repo because of the Node 10 requirement

package.json
```diff
@@ -25,7 +25,7 @@
     "eslint-plugin-jsdoc": "^30.5.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.1.4",
-    "husky": "^4.3.0",
+    "husky": "6.0.0",
```
alexander-akait, line 28: Please use `^` to get bug fixes and new features

nschonni: I'll update it, but there are no other releases to 6. I believe it was  a temporary one that works without the engine strict of Node 12 v7

nschonni: I've updated the range

sokra:
Thanks

Issue #316, opened 2021-11-01
Use throwIfNoEntry for sync resolver once graceful-fs supports it


markjm:
For cases of using sync resolver, throwIfNoEntry improves stat fail case by a large margin. The API in node is backwards compatible, so we *should* just be able to add this option and be on our way.

Unfortunately, there is an issue with this option in graceful-fs, so I have opened an issue there. Once resolved, this should be unblocked and easy change here.

https://github.com/isaacs/node-graceful-fs/issues/221

```
const fs = require("fs");
function oldStat(path) {
  let stat;

  try {
    stat = fs.statSync(path);
  } catch (e) {
    if (!(e && (e.code === "ENOENT" || e.code === "ENOTDIR"))) {
      throw e;
    }
  }
}

console.time("oldStat");
for (let i = 0; i < 100000; i++) oldStat("not-real-path");
console.timeEnd("oldStat"); // 6.7s

function newStat(path) {
  let stat = fs.statSync(path, { throwIfNoEntry: false });

}

console.time("newStat");
for (let i = 0; i < 100000; i++) newStat("not-real-path");
console.timeEnd("newStat"); // 2.1s
```

alexander-akait:
Make sense, PR welcome

PR #324, opened 2022-01-06
perf: utilize throwIfNoEntry in sync mode

markjm:
Resolves https://github.com/webpack/enhanced-resolve/issues/316

Additionally fixing a few types and updating graceful-fs to support the new option. Thankfully, `stat` being null (vs throwing) is already being handled correctly.

cc @alexander-akait - graceful-fs has been updated now so this is unblocked

### DIFF
--- a/lib/DirectoryExistsPlugin.js
+++ b/lib/DirectoryExistsPlugin.js
@@ -32,7 +32,7 @@ module.exports = class DirectoryExistsPlugin {
 					const fs = resolver.fileSystem;
 					const directory = request.path;
 					if (!directory) return callback();
-					fs.stat(directory, (err, stat) => {
+					fs.stat(directory, { throwIfNoEntry: false }, (err, stat) => {
 						if (err || !stat) {
 							if (resolveContext.missingDependencies)
 								resolveContext.missingDependencies.add(directory);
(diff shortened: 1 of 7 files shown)


markjm:
Actually - need to wait on this to figure out how to upgrade graceful-fs in webpack in a non-breaking way

ehoogeveen-medweb:
Was this blocked on https://github.com/webpack/webpack/pull/15118? If so, are all blockers resolved now?

markjm:
Yes, all blockers are resolved - the only concern would be if folks are using an older 4.xx version of webpack which doesnt contain the change you linked, but only upgraded enhanced-resolve. Any new installations of webpack would pick up this version, which is fine, but a repo with locked webpack version which tries to upgrade this would see some breakage.

That may be so niche of a break that its not worth considering, but I would defer to @sokra to decide how to integrate these changes.

Further note - this version works just fine "standalone", the potential issue comes because webpack passes in its own fs, and earlier versions of webpack fs contains a bug related to this setting.

sokra:
> Any new installations of webpack would pick up this version, which is fine, but a repo with locked webpack version which tries to upgrade this would see some breakage.

In the unlikely event of upgrading only enhanced-resolve, but not graceful-fs or webpack, we can expect people to update graceful-fs to fix the bug they are running into.

markjm:
Thanks @sokra - mind publishing this version?

PR #326, opened 2022-02-04
add yield in resolve context

vankop:
- If `resolveContext.yield` presented will return resolve results there `resolveContext.yield(obj)`.
- In this case resolve callback will not return any result.
- If no result yielded error with log will be returned as always.

### DIFF
--- a/lib/getPaths.js
+++ b/lib/getPaths.js
@@ -6,6 +6,7 @@
 "use strict";
 
 module.exports = function getPaths(path) {
+	if (path === "/") return { paths: ["/"], seqments: [""] };
 	const parts = path.split(/(.*?[\\/]+)/);
 	const paths = [path];
 	const seqments = [parts[parts.length - 1]];
--- a/test/yield.js
+++ b/test/yield.js
@@ -1,0 +1,169 @@
+const should = require("should");
+
+const { ResolverFactory } = require("../");
+const { Volume } = require("memfs");
+
+/** @typedef {import("../lib/Resolver").ResolveContext} ResolveContext */
+/** @typedef {ResolveContext & Required<Pick<ResolveContext, 'yield' | 'fileDependencies' | 'contextDependencies' | 'missingDependencies'>>} StrictResolveContext */
+
+const fileSystem = Volume.fromJSON(
+	{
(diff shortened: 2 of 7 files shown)


lib/getPaths.js
```diff
@@ -6,6 +6,7 @@
 "use strict";
 
 module.exports = function getPaths(path) {
+	if (path === "/") return { paths: ["/"], seqments: [""] };
```
vankop, line 9: have find a small bug where several same results were returned.
without this line `{ paths: ["/", "/"] }` returned

vankop:
maybe in `yield` case we need to use `forEachAll` instead of `forEachBail`, not sure that this make sense since anyway this improves only edge case performance.

test/yield.js
```diff
@@ -0,0 +1,169 @@
...
+			should(Array.from(fileDependencies.values()).sort()).be.eql([
+				"/",
+				"/a",
+				"/a/foo",
+				"/a/foo-2",
+				"/a/foo-2/b",
+				"/a/foo/b"
+			]);
+			should(Array.from(missingDependencies.values()).sort()).be.eql([
+				"/a/foo-2/b",
```
sokra, line 62: Interesting that this is in `missingDependencies` and in `fileDependencies`.
I guess it tries "as file" first and "as directory" second, resulting in it exists as file but is missing as directory...
Doesn't seem to be directly caused by this PR, so probably not relavant.

test/yield.js
```diff
@@ -0,0 +1,169 @@
...
+			should(Array.from(fileDependencies.values()).sort()).be.eql([
+				"/",
+				"/a",
+				"/a/foo",
+				"/a/foo/a",
+				"/b",
+				"/b/foo",
+				"/b/foo/a"
+			]);
+			should(Array.from(missingDependencies.values()).sort()).be.eql([
```
sokra, line 103:
```suggestion
			should(Array.from(missingDependencies).sort()).be.eql([
```

Technically that `.values()` is not necessary, but nevermind.

sokra (approved):

sokra:
Thanks

PR #327, opened 2022-02-07
chore: add cSpell commit hook

nschonni:
Since it is now being run on all files, it can be run as part of the pre-commit hook to check locally


alexander-akait:
@nschonni sorry for delay, can you rebase?

package.json
```diff
@@ -62,6 +62,7 @@
     "prepare": "husky install"
   },
   "lint-staged": {
+    "*": "cspell",
```
alexander-akait, line 65: I think we need `--no-must-find-files` here

nschonni:
```suggestion
    "*": "cspell --no-must-find-files",
```

alexander-akait (approved):

alexander-akait:
Thank you

PR #329, opened 2022-02-10
Fix argument name in error message

discordier:
The error message was referring to the `path` argument being not a string while it was the `request` argument being not a string.


alexander-akait (approved):

sokra:
Thanks

PR #331, opened 2022-02-17
fix yield

vankop:
- handle in ignore for AliasPlugin
- handle in ignore for AliasFieldPlugin
- handle for unsafe cache
- handle yield for custom plugins
- use real fs in yield tests
- add tests for AliasField and Alias in yield

### DIFF
--- a/lib/AliasFieldPlugin.js
+++ b/lib/AliasFieldPlugin.js
@@ -60,6 +60,10 @@ module.exports = class AliasFieldPlugin {
 						...request,
 						path: false
 					};
+					if (typeof resolveContext.yield === "function") {
+						resolveContext.yield(ignoreObj);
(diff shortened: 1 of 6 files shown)


sokra (commented):
I'm unsure if enhanced-resolve should do that special handling about the `path === false` object...

or if webpack should take care of that...

---

Why is an object with `path: false` special in enhanced-resolve?

If I ask enhanced-resolve to yield all results of an resolving, why doesn't it answer with the `path: false` object as one possible result, along with the others?

The special logic also breaks the assumption that the first yielded result is equal to the result without yield.

---

I would say that the special `path: false` object handling should be removed from enhanced-resolve... What do you think?

sokra (approved):

sokra:
Thanks

PR #336, opened 2022-03-18
chore: replace deprecated String.prototype.substr()

CommanderRoot:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### Breaking Changes

None

### DIFF
--- a/lib/AliasPlugin.js
+++ b/lib/AliasPlugin.js
@@ -48,7 +48,7 @@ module.exports = class AliasPlugin {
 										: resolver.join(item.name, "_").slice(0, -1)
 								))
 						) {
-							const remainingRequest = innerRequest.substr(item.name.length);
+							const remainingRequest = innerRequest.slice(item.name.length);
 							const resolveWithAlias = (alias, callback) => {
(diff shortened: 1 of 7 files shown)


alexander-akait (approved):

TheLarkInn:
@CommanderRoot congats on your first time contribution to enhanced-resolve! 🎉  Thank you for the PR.

PR #337, opened 2022-03-19
fix hash aliases

### LINKED ISSUE #334: About the new change in AliasPlugin.js
last week, your team update the version of the enhanced-resolve
however, using v5.9.2, my project cannot build at all, it will be stuck while building. 
It makes me really puzzled, because my project's alias is `'#': path.resolve(__dirname, './src'),`
when I change it into `'balabala': path.resolve(__dirname, './src'),`, it will continue to build. balabala can be anything but #
and when I change the line 47 into "request.path",  `'#': path.resolve(__dirname, './src'),` can also make it.
So I'm wondering the reason why your team make such a change

vankop:
fixes #334

### DIFF
--- a/lib/AliasPlugin.js
+++ b/lib/AliasPlugin.js
@@ -6,6 +6,7 @@
 "use strict";
 
 const forEachBail = require("./forEachBail");
+const { PathType, getType } = require("./util/path");
 
 /** @typedef {import("./Resolver")} Resolver */
 /** @typedef {import("./Resolver").ResolveRequest} ResolveRequest */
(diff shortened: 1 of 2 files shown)


test/alias.js
```diff
@@ -39,6 +39,9 @@ describe("alias", function () {
 				recursive: "recursive/dir",
 				"/d/dir": "/c/dir",
 				"/d/index.js": "/c/index",
+				// alias configuration should work
+				"#": "/c/dir",
+				"@": "/c/dir",
```
sokra, line 44: maybe also add a test case below for that

vankop: technically it fails without applying.. (just setting crash resolving) lets add usage anyway.

lib/AliasPlugin.js
```diff
@@ -30,6 +31,25 @@ module.exports = class AliasPlugin {
 	 */
 	apply(resolver) {
 		const target = resolver.ensureHook(this.target);
+		const cache = {};
```
sokra, line 34: The cache looks like an overkill for me here... Does it really make sense here?

sokra (changes requested):

lib/AliasPlugin.js
vankop: hm.. since mostly we won't have absolute aliases on average algorithm complexity will be `O(1)`.. lets remove it.

sokra (approved):

PR #347, opened 2022-05-23
fix imports field

### LINKED ISSUE #346: [Question] The behavior of importsField conflicts with `require.resolve`
For the [out of package scope of importField](https://github.com/webpack/enhanced-resolve/blob/main/test/importsField.js#L1214) example, there are some inconsistencies between `require.resolve` and `enhanced-resolve`.

Executing `require.resolve('#b')` in the importsField folder will report an error directly:

image

> node version: 12.22.10

But the above test case is executed normally.

So, should we add [checkTarget](https://github.com/webpack/enhanced-resolve/blob/main/lib/util/path.js#L195) logic to importsField as well?

vankop:
- throw error in imports field if trying to access out of package scope
- fix utility function `checkImportsExportsFieldTarget`

fixes #346

### DIFF
--- a/lib/ExportsFieldPlugin.js
+++ b/lib/ExportsFieldPlugin.js
@@ -10,7 +10,7 @@ const DescriptionFileUtils = require("./DescriptionFileUtils");
 const forEachBail = require("./forEachBail");
 const { processExportsField } = require("./util/entrypoints");
 const { parseIdentifier } = require("./util/identifier");
-const { checkExportsFieldTarget } = require("./util/path");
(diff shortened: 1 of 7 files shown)


bvanjoi:
maybe this case should be throw Error? https://github.com/webpack/enhanced-resolve/blob/main/test/importsField.js#L1214

vankop:
@bvanjoi nice catch! utility function had bug, I'll fix it..

sokra (approved):

sokra:
Thanks

PR #301, opened 2021-09-02
fix: find the `pnpapi` the `issuer` belongs to

merceyz:
**What's the problem this PR addresses?**

- In some cases `enhanced-resolve` is outside the control of the `pnpapi` and can't require it
- The `issuer` might not belong to the same `pnpapi` as `enhanced-resolve`
- The `issuer` might not belong to a `pnpapi` at all and should use the `node_modules` resolution instead

Fixes https://github.com/arcanis/pnp-webpack-plugin/issues/32 for Webpack 5

**How did you fix it?**

Use `module.findPnpApi` to locate the correct `pnpapi` for the `issuer`, if a `pnpapi` isn't found continue the normal resolution

PR #349, opened 2022-06-14
Fix `Cannot find module 'pnpapi'`

louisscruz:
As discussed in https://github.com/webpack/enhanced-resolve/issues/263, there are some scenarios in which `pnpapi` is not currently resolvable (e.g. `Cannot find module 'pnpapi'`). This happens when using `jest` to execute tests that make use of `webpack`. Here is [an example stack trace](https://github.com/webpack/enhanced-resolve/issues/263#issuecomment-1061189768), and [here is a repo for reproduction](https://github.com/webpack/enhanced-resolve/issues/263#issuecomment-1079894555).

It's odd that `pnpapi` is not resolvable when the [Yarn documentation mentions that it should be resolvable if `process.versions.pnp` is set](https://yarnpkg.com/advanced/pnpapi#processversionspnp). ~I've tried various other ways of loading up `pnpapi` based on @arcanis' [comment here](https://github.com/yarnpkg/berry/issues/693#issuecomment-575389842) without any luck. For some reason, `require('module').pnpApiPath` is `undefined`.~ Actually, it _does_ work if you resort to `require("module").findPnpApi(process.cwd());`! I previously thought I'd tried that, but must have gotten something wrong when trying.

### DIFF
--- a/lib/ResolverFactory.js
+++ b/lib/ResolverFactory.js
@@ -117,8 +117,13 @@ function processPnpApiOption(option) {
 		option === undefined &&
 		/** @type {NodeJS.ProcessVersions & {pnp: string}} */ versions.pnp
 	) {
-		// @ts-ignore
-		return require("pnpapi"); // eslint-disable-line node/no-missing-require
+		try {
+			// @ts-ignore
+			return require("pnpapi"); // eslint-disable-line node/no-missing-require
+		} catch (e) {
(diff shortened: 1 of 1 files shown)


louisscruz:
One thing this fails to take into account is the issuer of a given request. Having looked around a little more, I now see that the superior solution of https://github.com/webpack/enhanced-resolve/pull/301 already exists. Closing in favor of that.

Issue #352, opened 2022-07-25
Exports subpath wildcard patterns issue.


scott-lc:
Referring to the Node [subpath patterns](https://nodejs.org/api/packages.html#subpath-patterns) documentation, notice how a pattern can specify a file extension (in this example `.js`) on both sides of the pattern mapping:

```
{
  "exports": {
    "./features/*.js": "./src/features/*.js"
  }
}
```

According to the Node docs,

> Including the "*.js" on both sides of the mapping restricts the exposed package exports to only JS files.

However, this does not work in `enhanced-resolved`.   The wildcard pattern without a file extension works fine, for example:

```
{
  "exports": {
    "./features/*": "./src/features/*"
  }
}
```

But adding a wildcard file extension does not work.  To test this, simply add the following test case to `test/exportFields.js`:

```javascript
// ./test/exportFields.js
const testCases = [
  //...
  {
    name: "This will fail!",
    expect: ["./src/features/file.js"],
    suite: [
	{
	  "./features/*.js": "./src/features/*.js"
	},
	"./features/file.js",
	[]
    ]
  }
```
```

alexander-akait:
Sounds like a bug, do you want send a fix?

alexander-akait:
Fixed by https://github.com/webpack/enhanced-resolve/pull/353

PR #353, opened 2022-07-26
support wildcards pattern with common suffix in imports/exports field

bvanjoi:
solve https://github.com/webpack/enhanced-resolve/issues/352

### DIFF
--- a/lib/util/entrypoints.js
+++ b/lib/util/entrypoints.js
@@ -430,7 +430,24 @@ function targetMapping(
 		return mappingTarget + remainingRequest;
 	}
 	assert(mappingTarget, false);
-	return mappingTarget.replace(/\*/g, remainingRequest.replace(/\$/g, "$$"));
+	const wildcardIndex = mappingTarget.indexOf("*");
+
+	if (wildcardIndex !== -1) {
+		const maybeCommonSuffix = mappingTarget.slice(wildcardIndex + 1);
+		if (remainingRequest.endsWith(maybeCommonSuffix)) {
+			return mappingTarget.replace(
+				new RegExp(`\\*${maybeCommonSuffix}`, "g"),
+				remainingRequest.replace(/\$/g, "$$")
+			);
+		} else {
+			return mappingTarget.replace(
+				/\*/g,
+				remainingRequest.replace(/\$/g, "$$")
+			);
+		}
+	} else {
+		return mappingTarget;
+	}
 }
 
 /**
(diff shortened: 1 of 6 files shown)


lib/util/entrypoints.js
```diff
@@ -430,7 +430,24 @@ function targetMapping(
...
 	}
 	assert(mappingTarget, false);
-	return mappingTarget.replace(/\*/g, remainingRequest.replace(/\$/g, "$$"));
+	const wildcardIndex = mappingTarget.indexOf("*");
+
+	if (wildcardIndex !== -1) {
+		const maybeCommonSuffix = mappingTarget.slice(wildcardIndex + 1);
+		if (remainingRequest.endsWith(maybeCommonSuffix)) {
+			return mappingTarget.replace(
+				new RegExp(`\\*${maybeCommonSuffix}`, "g"),
```
dummdidumm, line 439: I think this needs to escape the characters that could have special meaning to the regex (like `.`)

dummdidumm (commented):
An alternative (which would require more work, or more significant changes, probably) would be to use [resolve.exports](https://www.npmjs.com/package/resolve.exports) which aims to consolidate resolution algorithms across tooling for this case.

lib/util/entrypoints.js
```diff
@@ -549,9 +566,10 @@ function walkPath(root, path, target) {
 		node.folder = target;
 	} else {
 		const file = lastNonSlashIndex > 0 ? path.slice(lastNonSlashIndex) : path;
-		if (file.endsWith("*")) {
+		const wildcardsIndex = file.indexOf("*");
```
alexander-akait, line 569: Weak check, in some filesystem file can have `*` character in the filename, let's use more strick checks, i.e. `*` or `*.js`

lib/util/entrypoints.js
alexander-akait: There is potential problem, we should handle only `*.js`, not `*` characters

alexander-akait (changes requested):
Let's use more stricky checks

alexander-akait:
And sorry for delay, @bvanjoi do you want to finish this PR
@dummdidumm you can resend this PR if we won't get an answer from @bvanjoi becasei it is so old :disappointed: sorry

bvanjoi:
> And sorry for delay, @bvanjoi do you want to finish this PR

I can refine this pr this weekend (forgive me for being a bit busy at work)

alexander-akait:
@bvanjoi Don't worry feel free to ping me

alexander-akait:
Almost finished, need to more test cases - WIP

alexander-akait (approved):

TheLarkInn (approved):

PR #357, opened 2022-10-15
fix: improve typescript defs

### LINKED ISSUE #356: Improve the typescript definitions
Currently many of the exported types are `any` or very weakly typed.

It would be great if these types could be properly populated via the jsdoc.

For example, the various places you can pass options tend to accept `any` even though we have a strong type for the user options already.

---

I did actually start doing this but the overloads you have lead to a lot of craziness in the types. 

For example, if we strongly type the overloads of `resolve`, the tests fail to compile because there are some which explicitly pass in `undefined` to ensure an error is thrown rather than bailing.

These tests make perfect sense but mean we end up having to dumb down the types to signatures nobody should ever really use. Alternatively we can double-cast undefined via jsdoc (to `unknown` then to `string`).

Similarly, the various overloads lead to all sorts of funky types, but are ultimately doable at least. 

Maybe if i at least put a draft pr up, you can see what i mean.

cc @sokra

43081j:
fixes #356 

i need some help with the overloads here i think.

you can see the strong types now mean a bunch of tests need to cast things too... im not sure there's a clean way to type the callback function strongly and not have to do those casts. the compiler is right: the result could be undefined if there's an error.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,6 +12,7 @@ const ResolverFactory = require("./ResolverFactory");
 /** @typedef {import("./PnpPlugin").PnpApiImpl} PnpApi */
 /** @typedef {import("./Resolver")} Resolver */
 /** @typedef {import("./Resolver").FileSystem} FileSystem */
+/** @typedef {import("./Resolver").ResolveCallback} ResolveCallback */
 /** @typedef {import("./Resolver").ResolveContext} ResolveContext */
 /** @typedef {import("./Resolver").ResolveRequest} ResolveRequest */
 /** @typedef {import("./ResolverFactory").Plugin} Plugin */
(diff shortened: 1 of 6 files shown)


lib/index.js
```diff
@@ -28,41 +29,86 @@ const asyncResolver = ResolverFactory.createResolver({
...
+			context = nodeContext;
+		}
+		return syncResolver.resolveSync(
+			context,
+			path,
+			/** @type {string} */ (request)
+		);
+	};
+
+/** @typedef {Omit<ResolveOptions, "fileSystem"> & Partial<Pick<ResolveOptions, "fileSystem">>} ResolveOptionsOptionalFS */
```
43081j, line 101: this thing bothers me. any suggestions welcome...

basically the `create` functions need to accept `ResolveOptions` with an optional filesystem, but everything else that takes options _requires_ a filesystem (the resolver factory etc)

lib/index.js
```diff
@@ -78,13 +124,19 @@ function create(options) {
 	};
 }
 
+/**
+ * @param {ResolveOptionsOptionalFS} options Resolver options
+ * @returns {{
+ *	(context: object, path: string, request: string): string|false;
+ *	(path: string, request: string): string|false;
+ * }} Resolver function
+ */
```
alexander-akait, line 133: Why do not move it to own type (same for other)?

43081j: yeah i agree. this one is specific to the synchronous resolver but i can extract it into a typedef and do the same for the async one

alexander-akait: Yeah, let's do it

43081j:
@alexander-akait have updated with the types we talked about, mind taking another look?

would be good if someone can trigger the CI too (first-time contributor 😬 )

43081j:
@sokra any chance you can take a look at this? or any other maintainer

alexander-akait (approved):

TheLarkInn (approved):
These are a significant improvement to enhanced-resolve. Thank you so much @43081j

PR #359, opened 2022-10-20
chore: update readme to reflect conditionNames default

### LINKED ISSUE #358: conditionNames defaults to `["node"]` according to docs, but possibly doesn't actually
the readme specifies:

https://github.com/webpack/enhanced-resolve/blob/651e57863e46afeccf1946b4275d7b2339152ed1/README.md?plain=1#L86

however:

https://github.com/webpack/enhanced-resolve/blob/651e57863e46afeccf1946b4275d7b2339152ed1/lib/ResolverFactory.js#L194

seems here we don't actually default the condition names array to anything, so later on the various plugins receive and empty set.

should it possibly be this instead:

```ts
conditionNames: new Set(options.conditionNames || ['node']),
```

43081j:
Fixes #358  by just documenting the actual default, rather than introducing a breaking change


alexander-akait (approved):

43081j:
@alexander-akait what do we need to do to merge this?

alexander-akait:
@sokra Friendly ping

PR #361, opened 2022-11-19
add vscode recommend config

fu1996:
When I studied the code base, I found that the development experience was not very good, so in order to have a unified and consistent development experience, I added the configuration related to vscode.


alexander-akait (commented):
Sorry, we don't ned it

PR #379, opened 2023-05-06
feat: migrate to jest

snitin315:
TODO:

- [x] Fix failing tests
- [x] Remove mocha
- [x] Remove Should


alexander-akait:
Looks very good, did you look why coverage is now only `81%`? I think we have a lot of tests so our coverage should be more...

snitin315:
@alexander-akait Yes, I'm migrating tests one by one. Once all the tests move to jest, coverage will be more.

package.json
```diff
@@ -27,12 +27,10 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.1.4",
     "husky": "^6.0.0",
+    "jest": "^27.5.1",
```
snitin315, line 30: Using jest v27 because it is compatible with node v10. We can upgrade it later.

snitin315:
@alexander-akait I have migrated all the tests. I think this is ready.

alexander-akait (approved):

PR #388, opened 2023-06-06
[draft] feat: resolve style field in package json

burhanuday:
Part of https://github.com/webpack/webpack/issues/14893

> supports style resolution in package.json and test it + test support loader usage

### DIFF
--- a/lib/StyleFieldPlugin.js
+++ b/lib/StyleFieldPlugin.js
@@ -1,0 +1,95 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Burhanuddin Udaipurwala @burhanuday
+*/
+
+"use strict";
+
+const path = require("path");
+const DescriptionFileUtils = require("./DescriptionFileUtils");
+
+/** @typedef {import("./Resolver")} Resolver */
+/** @typedef {import("./Resolver").JsonObject} JsonObject */
+/** @typedef {import("./Resolver").ResolveRequest} ResolveRequest */
+/** @typedef {import("./Resolver").ResolveStepHook} ResolveStepHook */
+
+const alreadyTriedStyleField = Symbol("alreadyTriedStyleField");
+
+module.exports = class StyleFieldPlugin {
+	/**
+	 * @param {string | ResolveStepHook} source source
+	 * @param {string} name fieldName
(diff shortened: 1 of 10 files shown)


lib/StyleFieldPlugin.js
```diff
@@ -0,0 +1,95 @@
...
+				if (
+					!stylePath ||
+					typeof stylePath !== "string" ||
+					stylePath === "." ||
+					stylePath === "./"
+				) {
+					return callback();
+				}
+				if (
+					// this.options.forceRelative
```
burhanuday, line 73: todo: add support for `forceRelative`

lib/StyleFieldPlugin.js
```diff
@@ -0,0 +1,95 @@
...
+		resolver
+			.getHook(this.source)
+			.tapAsync("StyleFieldPlugin", (request, resolveContext, callback) => {
+				// if request.context.issuer is not a css file, don't resolve to style field
+				const isIssuerAStyleFile =
+					request.context &&
+					// @ts-ignore
+					request.context.issuer &&
+					// @ts-ignore
+					request.context.issuer.endsWith(".css");
```
burhanuday, line 45: todo: check for all known css extensions
todo2: allow consumer to configure extensions array

lib/StyleFieldPlugin.js
```diff
@@ -0,0 +1,95 @@
...
+	apply(resolver) {
+		const target = resolver.ensureHook(this.target);
+		resolver
+			.getHook(this.source)
+			.tapAsync("StyleFieldPlugin", (request, resolveContext, callback) => {
+				// if request.context.issuer is not a css file, don't resolve to style field
+				const isIssuerAStyleFile =
+					request.context &&
+					// @ts-ignore
+					request.context.issuer &&
```
burhanuday, line 43: todo: check if type can be improved

alexander-akait:
@burhanuday It should not implemented here, we already support it in css-loader https://github.com/webpack-contrib/css-loader/blob/master/src/plugins/postcss-import-parser.js#L228, we just need to update defaults in webpack

burhanuday:
Ah. Will update the defaults in webpack. Did not occur to me that we could just change the default options. Closing this PR

PR #393, opened 2023-08-18
Reduce memory footprint of storing fs erros in cache

shYkiSto:
We are using `enhanced-resolve` together with `jest` test runner, and noticed substantial memory leaks using this combo. Further inspection revealed that original FS errors are retaining references to sandbox environments that jest creates to keep individual tests isolated from one another. These references occur in the internal representation of the `Error` object for lazy-computation of the `stack` property later in the process (when user code happens to read from it for the first time). 

Cloning the original error allows GC to collect that stack trace information and all retained objects that in turn helps to resolve the memory leak, and reduce the memory usage. One drawback to this is that the `stack` property would have to be computed eagerly, which comes at a cost of consuming more CPU. But since it's rarely required, and gets messed up due caching aspect of it anyway - a trade off has to be made to omit stack traces from the fs errors altogether.

### DIFF
--- a/lib/CachedInputFileSystem.js
+++ b/lib/CachedInputFileSystem.js
@@ -368,6 +368,19 @@ class CacheBackend {
 	 */
 	_storeResult(path, err, result) {
 		if (this._data.has(path)) return;
+
+		if (err) {
+			// Clone error object to be stored in cache to allow for the original
+			// error, and its internal representation to be picked up by GC. Thus
+			// reducing the memory footprint of caching such results in memory.
+			// Also, omitting expensive calculation of the stacktrace as it gets
+			// ignored most of the time anyway
+			const lightErrorProps = Object.getOwnPropertyDescriptors(err);
+			delete lightErrorProps.stack;
+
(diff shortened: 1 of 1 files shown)


rtsao:
Any feedback @alexander-akait @TheLarkInn ? This would be nice fix to get landed

Twipped:
Is there a workaround for this that we can employee while we wait for this to be merged? We're seeing substantial memory problems in our jest suites that we've traced back to this exact issue.

alexander-akait:
Sorry for delay, it sounds like the problem is with jest...

shYkiSto:
> Sorry for delay, it sounds like the problem is with jest...

Jest's approach to sandboxing definitely presents a challenge here. But at the same time there's not much that can be done there, as it's essentially the user code (enhanced-resolve) causing the leakage by holding references to the test environments that prevents them from being GC'd, as summarized in the PR description.

alexander-akait:
@shYkiSto Sorry for the delay, can you measure the memory or demonstrate the leak (small reproduce example)?

Issue #401, opened 2024-01-03
Add ability to handle win32 relative paths


davakh:
### Issue
Right now, based on codebase, library supports only posix relative paths. 

### Reproduction
This is **correct** relative path in context of this library:
```
../../../src
```

This is **incorrect** relative path in context of this library:
```
..\\..\\..\\src
```

To understand how the library understands these paths, you can call `getType` from [this file](https://github.com/webpack/enhanced-resolve/blob/58464fc/lib/util/path.js#L40).

So, anything that can be passed as a result by [path.relative(...)](https://nodejs.org/api/path.html#pathrelativefrom-to) in Node.js, will not work as expected on Windows machines. 

### Solutions

Possible solution is to add separation in the enum of types for `RelativePosix` and `RelativeWin`, like for the `Absolute..` paths. But it's hard to predict what integrations with this library are going to break, so I think it can be a big deal. 

### Workaround

Add replacing for backward compatibility with Windows (Windows can handle posix separator):
```
path.relative(....).split(path.win32.sep).join(path.posix.sep);
```

----------

**PS.**  Maybe it's the only way you want to handle paths in this library, and then I think it would be cool to add information about this scenario or to close this issue with a preferable workaround.

alexander-akait:
Hello, can you provide a problem/example where you got such paths?

davakh:
[...]

PR #402, opened 2024-02-07
feat: add support for win32 paths

davakh:
I implemented general logic for resolving win32 relative paths in library. It's hard to say what else to cover with tests because it's hard to fully mock Node.js's path module and easily switch between win32/posix paths resolving in jest.

If you have thoughts on what else should I cover, I would like to hear about it and I can improve coverage if it's required.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node-version: [10.x, 12.x, 14.x, 16.x, 18.x, 20.x, 22.x]
+        node-version: [10.x, 12.x, 14.x, 16.x, 18.x, 20.x, 22.x, 23.x]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4


alexander-akait (commented):
Thank you looks good, can we add more test with real examples?

alexander-akait:
Also please fix ts problems, thank you

alexander-akait:
@davakh Hello, friendly ping :star:

davakh:
Hello, sorry that it took so long. I've added tests and rewritten the original `path` tests based on the fact that this repository has separate runners for Windows. Also, I'm still not sure what you mean by "real examples" and want to ask for a hint about that, but I've added platform-based tests to the main function, so it should cover more cases.

alexander-akait (approved):
There are enough tests. Thank you

davakh:
It's hard for me to debug on Windows, and it feels like it requires refactoring of almost all existing tests because they're not based on `path.sep` from the Node.js path package, and all the existing tests are using POSIX-char separators. 

I think it would have a greater impact with a lower cost and a lower chance of breaking something in conclusion if we are going to apply something like this `path.relative(....).split(path.win32.sep).join(path.posix.sep);` at the beginning of resolving path.

It is hard for me to decide because I don't like this solution. If you agree, I can add this solution to codebase, but if you have any other ideas about solving this, I would like to know about it.

alexander-akait:
Yeah, we need refactor our tests

davakh:
I [refactored existing tests](https://github.com/davakh/enhanced-resolve/commit/3cdeb7af42827d7626487666145d4528c64438dc), but there are more things to think about:
1. `memfs` is partially supporting win32 paths, you can't create `C:\` volume files or any absolute win32 path to test it against  current tests in `enhanced-resolve`.  So, if we want to correctly run tests for Windows, we should require changes from `memfs` or switch to another library. This is huge change for existing tests. **Shorter:** it's impossible to create volume in `memfs` and write/read with absolute win32 paths
 2. Also, there's an issue with paths that can later be resolved, but rn it feels like developers of node.js just want to keep it like this because many project can relate to this behavior. Related issue: https://github.com/nodejs/node/pull/51513 . It means sometimes win32 paths can be referenced to UNC (Universal Naming Convention) paths rather than actual paths because of [this](https://github.com/webpack/enhanced-resolve/blob/a69818341516d2800abef8f454aa1b79ca4a4122/lib/util/path.js#L167) concatenation before normalization. i.e. `path.normalize('//./a') # /./a/` but on win32 paths it's `path.normalize('\\\\.\\a') # \\\\.\\a\\`. **Shorter**: posix node.js path module can't handle UNC paths, but it handles UNC paths correctly on win32. 

Current Node.js behavior with UNC paths:
```
> path.normalize('\\\\.\\a')
'\\\\.\\a' # Works as expected
> path.normalize('//./a');
'/a' #  Doesn't recognize UNC paths
```
 
**Possible solutions:** 
1.  Replace or expand `memfs` library to use in tests. Maybe use another library exclusively for Windows tests.
2. [This line](https://github.com/webpack/enhanced-resolve/blob/a69818341516d2800abef8f454aa1b79ca4a4122/lib/util/path.js#L167) can be replaced to 
```
winNormalize(`${rootPath}${rootPath === '\\' ? '.\\' : '\\'}${request}\`);
``` 
to keep expected behavior - cause it feels more expected behavior to transfer `\\\\.\\a` -> `\\.\\a\\` rather than `\\\\a\\` (actual with win32 paths) based on usage of this function in `ehanced-resolve` library. 

> Reasoning behind this message: I'm still thinking about how to better resolve all these questions, but right now while I'm thinking about it, it feels like it can be helpful to someone if I let these thoughts exist at least here.

alexander-akait:
@davakh 

 > Replace or expand memfs library to use in tests. Maybe use another library exclusively for Windows tests.
 
 Maybe we can fake the behavior in places we need it just for testing?

davakh:
> > Replace or expand memfs library to use in tests. Maybe use another library exclusively for Windows tests.
> 
> Maybe we can fake the behavior in places we need it just for testing?

Okay, I dug deeper and with current set of methods required to use by fs in library, `memfs` works correctly for most cases. **We can keep `memfs` like this with a few changes to the tests.**

--- 
 
**About current progress:** 
I investigate how Node.js works with exports/imports fields of `package.json` and how I can apply win32 paths to existing functionality of the library. It's not obvious because it requires from package.json paths to start with posix-slash `./` at least [by the Node.js documentation](https://nodejs.org/api/packages.html#exports):
> All paths defined in the "exports" must be relative file URLs starting with `./`.

Because of this requirement to exports/imports fields, I have a lot of thoughts about returning to another option: modify win32 path at the beginning of any library public api and work in the library only with posix slash (`/`) paths even on Windows because path.win32 also can work with the posix slash.

alexander-akait:
> Because of this requirement to exports/imports fields, I have a lot of thoughts about returning to another option: modify win32 path at the beginning of any library public api and work in the library only with posix slash (/) paths even on Windows because path.win32 also can work with the posix slash.

Sounds good for me - less changes in code

alexander-akait:
I think we can solve this - https://github.com/webpack/enhanced-resolve/issues/430 too, using logic above

davakh:
Okay, sorry, I didn't expect for PR to be closed on force-push.. 
Related pull request: https://github.com/webpack/enhanced-resolve/pull/436

PR #418, opened 2024-04-30
pathUtils | fix memory leak

Galcarmi:


package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "graceful-fs": "^4.2.4",
+    "lru-cache": "^10.2.2",
```
alexander-akait, line 17: I don't think we need this big package here, you can rewrite this package and put it inside project, because we use only the `max` value

Galcarmi: i'll implement it and update

lib/util/path.js
```diff
@@ -6,6 +6,7 @@
 "use strict";
 
 const path = require("path");
+const { LRUCache } = require("lru-cache");
```
alexander-akait, line 9: Let's use your implementation here and remove this package from here

alexander-akait (commented):
Also there is simple and fast example:
```
class LRU {
    constructor(max = 10) {
        this.max = max;
        this.cache = new Map();
    }

    get(key) {
        let item = this.cache.get(key);
        if (item !== undefined) {
            // refresh key
            this.cache.delete(key);
            this.cache.set(key, item);
        }
        return item;
    }

    set(key, val) {
        // refresh key
        if (this.cache.has(key)) this.cache.delete(key);
        // evict oldest
        else if (this.cache.size === this.max) this.cache.delete(this.first());
        this.cache.set(key, val);
    }

    first() {
        return this.cache.keys().next().value;
    }
}
```

alexander-akait:
@vankop What do you think about it? Yeah, we have a memory leak here, if you run develpment your application very long our cache will be bigger and bigger, so I think using lru cache is good idea, but I don't know what is will be better number for max

vankop:
I personally think we dont need cache here.. maybe at first implementation there was some specific usage for `resolver.join`, but right now it used in a lot of cases, sometimes even on each plugin call. ( see JoinPlugins ) So probably cache hit is very low. basically happen only on the same resolving, this happens quite rarely on incremental builds

Galcarmi:
@alexander-akait sorry I'm a bit busy, I'll address it at the weekend,

@vankop so do you recommend to just remove it instead?

vankop:
we have a metrics repo, we just need to setup cache hits there...

vankop:
btw I guess it is good idea

PR #429, opened 2024-07-18
fix: `exports` and `imports` array target resolving

alexander-akait:
fixes https://github.com/webpack/enhanced-resolve/issues/400


hardfist:
@alexander-akait does this deprecate https://webpack.js.org/guides/package-exports/#alternatives behavior? will this be shipped in webpack@5?

alexander-akait:
@hardfist we should rewrite our docs because we are align with Node.js behaviour right now

alexander-akait:
https://github.com/webpack/webpack.js.org/issues/7323

alexander-akait:
@hardfist Technically you are right, but I would say it is better to fix it in the project, because relying on behavior that was initially designed incorrectly is not the best idea, most likely it would break in other bundlers and in the node itself, as workarund you can fix old version or just copy paste our old code (as a plugin) and set other `stage`

PR #436, opened 2024-10-30
fix: corrects resolve for win32 paths

### LINKED ISSUE #401: Add ability to handle win32 relative paths

davakh:

accidentaly closed previous PR: https://github.com/webpack/enhanced-resolve/pull/402


test/pnp.test.js
```diff
@@ -112,15 +138,19 @@ describe("pnp", () => {
...
+			expect(result).toEqual(
+				transferPathToPosix(path.resolve(fixture, `pkg${obps}main.js`))
+			);
 			done();
 		});
 	});
 	it("should resolve namespaced module names", done => {
-		pnpApi.mocks.set("@user/pkg", path.resolve(fixture, "pkg"));
-		resolver.resolve({}, __dirname, "@user/pkg", {}, (err, result) => {
+		pnpApi.mocks.set(`@user${posixSep}pkg`, path.resolve(fixture, "pkg"));
```
davakh, line 148: thoughts: i'm not sure about using `posixSep` in this place, because it should use `osBasedPathSeparator (obps)` with the reasoning from `/test/util/path-separator`

test/util/path-separator.js
```diff
@@ -0,0 +1,35 @@
...
+// This function is used to transfer path from win32 to posix path separator. It usually used in tests to compare results. We can omit this by refactor all the existing tests
+// where a test use something like this "path.resolve(fixture, '/example/index.js')" code in expecting result. Right now, I'm solving another problem, so I'm not going to do it right now - just to make less
+// changes in existing tests and make this transition to Windows-friendly code easier.
+const transferPathToPosix = (path) => path.replace(/\\/g, '/');
+
+module.exports = {
+    absoluteOsBasedPath,
+    absoluteOsBasedResolvedPath,
+		posixSep: posixPathSeparator,
+		obps: osBasedPathSeparator,
```
davakh, line 33: thoughts: of course, I wanted to use fullname, but this entity being used too often in tests, so I decided to make it shorter

lib/util/win32-normalize-options.js
```diff
@@ -0,0 +1,108 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
```
davakh, line 2: question: not sure what i should write here, every other file has author, should I add same author (`Author Tobias Koppers @sokra`) as other files?

test/yield.test.js
```diff
@@ -3,42 +3,60 @@ const fs = require("fs");
...
+	obps
+} = require("./util/path-separator");
 
 /** @typedef {import("../lib/Resolver").ResolveContext} ResolveContext */
 /** @typedef {ResolveContext & Required<Pick<ResolveContext, 'yield' | 'fileDependencies' | 'contextDependencies' | 'missingDependencies'>>} StrictResolveContext */
 
+// Functions below sometime handle transferPathToPosix and based on reading of test file you can see, that
+// sometimes absolute paths started with obps (osBasedPathSeparator) and not absoluteOsBasedPath. That's because we use path.join too much in this test,
+// and it works tricky on platforms:
+// > path.posix.join('/abc/ab', '/abc/de') -> '/abc/ab/abc/de' (correct path after all, )
```
davakh, line 18: lint: I'll fix it with following changes after review

lib/util/win32-normalize-options.js
alexander-akait: @davakh You can write own name here

lib/ResolverFactory.js
```diff
@@ -199,7 +202,7 @@ function createOptions(options) {
 		}
 	}
 
-	return {
+	return normalizeOptionsForWindows({
```
alexander-akait, line 205: I am afraid it will be slow on big builds...

alexander-akait (commented):
Why just don't coonvert it to unix path and work internaly only with `/`? I just don't understand some of the changes.

lib/DescriptionFileUtils.js
```diff
@@ -189,6 +189,8 @@ function getField(content, field) {
  */
 function cdUp(directory) {
 	if (directory === "/") return null;
+	if (directory.match(/^[a-zA-Z]:$/)) return null;
```
alexander-akait, line 192: Why we need it? We have unix path internally...

lib/PnpPlugin.js
```diff
@@ -77,10 +79,18 @@ module.exports = class PnpPlugin {
 						return;
 					}
 
+					if (typeof resolution === "string") {
+						resolution = transformPathToPosix(resolution);
```
alexander-akait, line 83: Same here

lib/RestrictionsPlugin.js
```diff
@@ -9,7 +9,6 @@
 /** @typedef {import("./Resolver").ResolveStepHook} ResolveStepHook */
 
 const slashCode = "/".charCodeAt(0);
-const backslashCode = "\\".charCodeAt(0);
```
alexander-akait, line 12: Can you describe?

lib/getPaths.js
```diff
@@ -11,7 +11,7 @@
  */
 module.exports = function getPaths(path) {
 	if (path === "/") return { paths: ["/"], segments: [""] };
-	const parts = path.split(/(.*?[\\/]+)/);
+	const parts = path.split(/(.*?[/]+)/);
```
alexander-akait, line 14: And this

lib/ResolverFactory.js
davakh: Okay, I'll take a look how it integrates with other libraries, maybe I'll find better place to normalize it

lib/DescriptionFileUtils.js
davakh: Because it can be win32 path still. Of course, it has posix char separator, but absolute path can be resolved as `X:/something/` for example and for this case - to prevent this function to go up

davakh:
> Why just don't coonvert it to unix path and work internaly only with `/`? I just don't understand some of the changes.

That's what i did, that's how it works right now in this PR: we convert the external paths (we can get it in resolver or with options) and work internally only with `/`

lib/PnpPlugin.js
davakh: external path, that can be received from pnpapi -> we transform it to path with posix separators in case we get win32 separator from outside

lib/RestrictionsPlugin.js
davakh: We can keep it like this, but I thought about removing it because internally we should handle only posix separator char

lib/getPaths.js
davakh: Same idea as this: https://github.com/webpack/enhanced-resolve/pull/436#discussion_r1826329289

lib/util/path.js
```diff
@@ -201,3 +200,16 @@ const cachedJoin = (rootPath, request) => {
...
+ * @param {string} rawPath node.js path
+ * @returns {string} normalized for windows, same for posix paths
+ */
+const transformPathToPosix = rawPath => {
+	if (!rawPath) return rawPath;
+
+	if (path.sep === "/")
+		return rawPath; // Don't even transform if path.sep is posix
+	else return rawPath.split("\\").join("/");
+};
```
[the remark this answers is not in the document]
davakh: I agree about the performance part. About the behavior: it's already different and based on os.  

It's possible with this PR to test how different behavior between Unix and Windows with the following instruction:
- revert changes from library itself, keep the changes for tests
- mock `transferPathToPosix` function in `test/util/path-separator.js` because it exists in tests only for assertion and handles assertion for changes exclusively in current pull request (it removes the conversion from `\\` to `/` provided by changes in library which we reverted)

alexander-akait:
Close in favor of https://github.com/webpack/enhanced-resolve/pull/564

PR #438, opened 2024-12-10
optimize alias plugin for huge alias objects

VincentBailly:
The alias plugin gets very slow when the webpack config has many aliases.

This PR aims to re-implement the features of the current alias plugin but in a more efficient manner:

- Create only the minimum amount of closure objects as possible
- Memoize string operations by running them in the constructor of the plugin

### DIFF
--- a/lib/AliasPlugin.js
+++ b/lib/AliasPlugin.js
@@ -1,11 +1,10 @@
 /*
-	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
+  MIT License http://www.opensource.org/licenses/mit-license.php
+  Author Tobias Koppers @sokra
 */
 
 "use strict";
 
-const forEachBail = require("./forEachBail");
 const { PathType, getType } = require("./util/path");
 
 /** @typedef {import("./Resolver")} Resolver */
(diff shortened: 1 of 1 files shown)


lib/AliasPlugin.js
```diff
@@ -23,6 +22,9 @@ module.exports = class AliasPlugin {
 	constructor(source, options, target) {
 		this.source = source;
 		this.options = Array.isArray(options) ? options : [options];
+		for (const item of this.options) {
+			item.nameWithSlash = item.name + "/";
```
alexander-akait, line 26: Why we need it?

VincentBailly: It's just a memoization so we don't do this concatenation on every resolution, the concatenation creates an object that is short lived, creating work for the garbage collector. By creating the concatenated string once, storing it and reusing it we save the memory overhead of creating many times the same temporary objects.

alexander-akait:
Tests are failed

lib/AliasPlugin.js
dmichon-msft: If the aliasing object gets truly large, it'd be worth investing in adding a dependency on a prefix-tree implementation that can handle the optional `*` component in the path, and use that to optimize the matching logic.

dmichon-msft: To avoid shape changes it would be best to have `this.options` consist of fresh objects that have all required properties.

alexander-akait:
Fixed, now we faster around 2x-3x

PR #445, opened 2025-01-10
feat: monomorphic ResolveRequest object

dmichon-msft:
Improves performance all-up (especially in `ParsePlugin`, `JoinRequestPlugin`, `JoinRequestPartPlugin`) by ensuring that the `ResolveRequest` object always has a constant shape.

Additionally, creating the object via a direct object literal is _significantly_ faster than creating it via object spread.

Testing on NodeJS 18.19.1, this had the following impact for a local webpack build.
Before:
image
image

After:
image
image

### DIFF
--- a/lib/AliasFieldPlugin.js
+++ b/lib/AliasFieldPlugin.js
@@ -68,8 +68,25 @@ module.exports = class AliasFieldPlugin {
 				if (data === false) {
 					/** @type {ResolveRequest} */
 					const ignoreObj = {
-						...request,
-						path: false
+						path: false,
+						context: request.context,
+						descriptionFilePath: request.descriptionFilePath,
+						descriptionFileRoot: request.descriptionFileRoot,
+						descriptionFileData: request.descriptionFileData,
+						relativePath: request.relativePath,
+						ignoreSymlinks: request.ignoreSymlinks,
+						fullySpecified: request.fullySpecified,
+						__innerRequest: request.__innerRequest,
+						__innerRequest_request: request.__innerRequest_request,
+						__innerRequest_relativePath: request.__innerRequest_relativePath,
+
+						request: request.request,
+						query: request.query,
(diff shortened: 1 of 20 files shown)


dmichon-msft:
> @dmichon-msft Yeah, I am afraid it is a breaking change...
> 
> In webpack we have such things - https://github.com/webpack/webpack/blob/main/lib/cache/ResolverCachePlugin.js#L147 and https://github.com/webpack/webpack/blob/main/lib/cache/ResolverCachePlugin.js#L150
> 
> Technically we can solve this and add a special place to store such things.
> 
> I have only remembered one place where we use it, maybe somewhere else, but it is quite possible that this is the only place
> 
> So let's try to solve it step by step
> 
> 1. Implement special place to store custom values for authors (maybe just Map?)
> 2. Make internal helper for request generation and implement an option to back compat
> 3. Rewrite it in webpack and test it
> 
> Alternative solution, maybe even better, but need to check perf
> 
> 1. Check request in factory (we can use Symbol `isPure`) - if no unknown properties it is pure (we known them), otherwise `false`
> 2. Make internal helper for creating request - if it `isPure` request we can use your logic and get perf improvement, if no just use olg logic
> 3. (Optional) And also provide a place to store custom values for authors, because to bring such improvements to webpack
> 
> So we will keep old behaviour and have perf improvement for some places

The `context` property is already a completely opaque object that is forwarded through the request chain without being inspected or modified in any way, so we ought to be able to just use that for the ResolverCachePlugin scenario.

dmichon-msft:
I have some thoughts on ways to make this behavior be a resolver option, will refactor.

alexander-akait:
Yeah, an option is a good way here, also maybe we should add a place where we can store custom values, so after merge this we can update logic inside webpack and get perf (we need a place where to store custom values inside webpack for caches)

PR #456, opened 2025-08-05
fix: nonsensible intersection

colinaaa:
After upgrading to Webpack ^5.100.0 (https://github.com/lynx-family/lynx-stack/pull/1418), we got errors related with types:

```
- JsonObjectTypes[key]: string & Array<JsonValueTypes>
  - nonsensible intersection

- JsonObjectTypes[key]: number & Array<JsonValueTypes>
  - nonsensible intersection

- JsonObjectTypes[key]: false & Array<JsonValueTypes>
  - nonsensible intersection

- JsonObjectTypes[key]: true & Array<JsonValueTypes>
  - nonsensible intersection
```

See: https://github.com/lynx-family/lynx-stack/actions/runs/16744114403/job/47401398078

### DIFF
--- a/lib/AliasFieldPlugin.js
+++ b/lib/AliasFieldPlugin.js
@@ -44,9 +44,7 @@ module.exports = class AliasFieldPlugin {
 				if (fieldData === null || typeof fieldData !== "object") {
 					if (resolveContext.log) {
 						resolveContext.log(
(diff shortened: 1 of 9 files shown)


alexander-akait:
@colinaaa Looks like lint is failed

PR #463, opened 2025-09-11
feat: add TsconfigPathsPlugin

xiaoxiaojx:


lib/TsconfigPathsPlugin.js
```diff
@@ -0,0 +1,238 @@
...
+ * @property {string=} baseUrl - Override baseUrl from tsconfig.json
+ * @property {string=} context - Context directory for resolving relative paths
+ */
+
+module.exports = class TsconfigPathsPlugin {
+	/**
+	 * @param {string} configFile tsconfig file path
+	 */
+	constructor(configFile) {
+		this.configFile = configFile;
```
alexander-akait, line 28: Ideally we should accept just `true` option too (i.e. `string` and `true`, maybe better name for `true` is `auto`, what do you think?), it means try to load tsconfig, if not found do nothing, for better development experience, setting paths to the config file can takes a time and sometimes it's not very convenient

lib/TsconfigPathsPlugin.js
```diff
@@ -0,0 +1,238 @@
...
+				),
+			);
+
+			// Collect references from the main tsconfig.json
+			const tsconfigJson = await new Promise((resolve, reject) => {
+				fs.readFile(configPath, "utf8", (err, data) => {
+					if (err) reject(err);
+					resolve(JSON.parse(/** @type {string} */ (data)));
+				});
+			});
```
alexander-akait, line 124: We are reading configuration file on the each resolve, it can have a bad performance

lib/TsconfigPathsPlugin.js
```diff
@@ -0,0 +1,238 @@
...
+	 * @returns {Promise<AliasOption[]>} Array of alias options for AliasPlugin
+	 */
+	async loadAndConvertPaths(resolver) {
+		try {
+			const fs = resolver.fileSystem;
+			const configPath = path.isAbsolute(this.configFile)
+				? this.configFile
+				: resolver.join(process.cwd(), this.configFile);
+
+			const mainOptions = await this.readTsconfigCompilerOptions(
```
alexander-akait, line 102: I think we can cache `readTsconfigCompilerOptions` using (and the same for `readTsconfigCompilerOptions` below)
```js
this.tsconfigProcessorCache = new WeakMap<TS_CONFIG_JSON_VALUE, { aliases: Aliases, references: References }>;
```

So we will read everything only on the first run, and will keep results until GC allow it

Feel free to think how we can avoid more tasks on the second and third calls

lib/TsconfigPathsPlugin.js
```diff
@@ -0,0 +1,238 @@
...
+				...this.convertPathsToAliases(
+					resolver,
+					mainOptions.paths,
+					mainOptions.baseUrl,
+				),
+			);
+
+			// Collect references from the main tsconfig.json
+			const tsconfigJson = await new Promise((resolve, reject) => {
+				fs.readFile(configPath, "utf8", (err, data) => {
```
alexander-akait, line 120: Two read of the `configPath` - here and in `readTsconfigCompilerOptions`, need to think how refactor it to avoid it

lib/TsconfigPathsPlugin.js
xiaoxiaojx: the current implementation only reads tsconfig.json once during instantiation, so it already acts as a cache.
```js
apply(resolver) {
		 // 👇 it only reads once when the plugin is instantiated
		// similar to how this.tsconfigProcessorCache works.
		const aliasOptionsPromise = this.loadAndConvertPaths(resolver);

```
In most cases, TsconfigPathsPlugin will only be instantiated once, so aliasOptionsPromise acts as a cache that ensures it’s only read once.
In less common cases where the user calls ResolverFactory.create for every resolve, the plugin will be instantiated multiple times. In that case, if users want caching, they can enable it via the unsafeCache option.

xiaoxiaojx: What if we do it like package.json and attach the data to [request.descriptionFileData](https://github.com/webpack/enhanced-resolve/blob/main/lib/DescriptionFilePlugin.js#L49) 🤔?
If the property already exists, we just skip reading the file again.

in the latest implementation I attached it to [request.tsconfigFileData](https://github.com/webpack/enhanced-resolve/pull/463/files#diff-b7aef0969ef4cdabf7bce8e5bc46e2e60a774daa2d5581cf5e6447eab2e0c345R311).

xiaoxiaojx:
The current implementation doesn’t seem to handle cases where the paths value is a complex array.
I’ll switch the PR status to WIP.
```json
{
  "compilerOptions": {
    "paths": {
      "lib1": ["../foo", "./lib1", "./bar"],
      "lib2/*": ["../foo", "./lib2/*"],
      "*/lib3": ["../foo", "./lib3/*", "./bar"]
    }
  }
}
```

lib/TsconfigPathsUtils.js
```diff
@@ -0,0 +1,299 @@
...
+		let config;
+		try {
+			config = JSON.parse(/** @type {string} */ (data));
+		} catch (err) {
+			throw new Error(
+				`${configFilePath} is malformed ${/** @type {Error} */ (err).message}`,
+			);
+		}
+
+		const extendedConfig = config.extends;
```
alexander-akait, line 153: Good catch :+1:

lib/ResolverFactory.js
```diff
@@ -294,6 +297,7 @@ function createOptions(options) {
 		preferRelative: options.preferRelative || false,
 		preferAbsolute: options.preferAbsolute || false,
 		restrictions: new Set(options.restrictions),
+		tsconfig: typeof options.tsconfig === "undefined" ? true : options.tsconfig,
```
alexander-akait, line 300: By default it should be false, not the each project uses ts, so let's avoid extra job, developer should set `resolve.tsconfig: true | pathToTsconfig.json` to enable this plugin

lib/TsconfigPathsUtils.js
```diff
@@ -0,0 +1,299 @@
...
+) {
+	// Add .json extension if not present
+	if (
+		typeof extendedConfigValue === "string" &&
+		!extendedConfigValue.includes(".json")
+	) {
+		extendedConfigValue += ".json";
+	}
+
+	const currentDir = path.dirname(configFilePath);
```
alexander-akait, line 92: Can we implement `path.dirname` in `util/path` and cache it too to better perf?

lib/TsconfigPathsUtils.js
```diff
@@ -0,0 +1,299 @@
...
+		const result = await readTsconfigCompilerOptions(
+			fileSystem,
+			absTsconfigPath,
+		);
+		if (!result) return null;
+		const { options: compilerOptions, fileDependencies } = result;
+		if (!compilerOptions) return null;
+
+		return {
+			...tsconfigPathsToResolveOptions(
```
alexander-akait, line 288: We can cache this, look at `ExportsFieldPlugin.js` plugin, we have `this.fieldProcessorCache` to cache processed export fields, so let's use the same logic, just using `this.tsconfigProcessorCache` name and cache result of this function

alexander-akait:
Add let's add more test cases to improve coverage

lib/TsconfigPathsUtils.js
```diff
@@ -0,0 +1,299 @@
...
+	// Check if file exists, if not try node_modules
+	const exists = await new Promise((resolve) => {
+		fileSystem.readFile(extendedConfigPath, (err) => {
+			resolve(!err);
+		});
+	});
+	if (
+		!exists &&
+		extendedConfigValue.includes("/") &&
+		extendedConfigValue.includes(".")
```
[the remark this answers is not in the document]
xiaoxiaojx: The goal here is to resolve the extendedConfigValue path from node_modules, e.g. extends: "react/tsconfig.json". Reference: https://github.com/jonaskello/tsconfig-paths/blob/master/src/tsconfig-loader.ts#L187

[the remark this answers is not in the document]
xiaoxiaojx: Added a test case here https://github.com/webpack/enhanced-resolve/pull/463/files#diff-91224cf57750b2195ca9c17182d982488414303921bf1332a7e1081ed05953b1R2

lib/TsconfigPathsUtils.js
xiaoxiaojx: The tsconfigPathsToResolveOptions function only runs once per resolve. However, the key of this.fieldProcessorCache is JsonObject /** @type {JsonObject} */ (request.descriptionFileData), and since each resolve produces a different JsonObject reference, this cache becomes ineffective in scenarios where it should run only once per resolve. This makes me wonder whether the cache in ExportsFieldPlugin (this.fieldProcessorCache) aligns with the original intent—perhaps the key should be a string rather than an object reference.

README.md
```diff
@@ -111,6 +111,7 @@ myResolver.resolve(
 | restrictions     | []                          | A list of resolve restrictions                                                                                                                            |
 | roots            | []                          | A list of root paths                                                                                                                                      |
 | symlinks         | true                        | Whether to resolve symlinks to their symlinked location                                                                                                   |
+| tsconfig         | true                        | Path to tsconfig.json for paths mapping. Default true loads tsconfig.json, or pass a string path.                                                         |
```
alexander-akait: It should be `false` by default

lib/TsconfigPathsPlugin.js
```diff
@@ -0,0 +1,405 @@
...
+			return new Promise((resolve, reject) => {
+				readJson(jsonFilePath, (err, content) => {
+					if (err) return reject(err);
+					resolve(/** @type {Tsconfig} */ (content));
+				});
+			});
+		}
+
+		const content = await new Promise((resolve, reject) => {
+			fileSystem.readFile(jsonFilePath, "utf8", (err, data) => {
```
alexander-akait: Let's avoid using `utf8` for any reads for caching purposes

PR #526, opened 2026-04-16
perf: singly-linked StackEntry for resolver stack with Set-like API

xiaoxiaojx:
**Summary**
Replace `Set<string>`-based stack with a singly-linked `StackEntry` class that exposes a Set-compatible API (`has`, iteration, `size`, `toString`). Each `doResolve` call prepends a node instead of cloning a Set.

### Comparison

| | [original PR #443](https://github.com/webpack/enhanced-resolve/pull/443) | [PR #523 (`claude/pr-443-...`)](https://github.com/webpack/enhanced-resolve/pull/523) | **this PR (#526)** |
|---|---|---|---|
| Stack representation | POJO `{ name, path, request, ..., parent }` linked list | POJO `{ entry, parent }` linked list, hidden behind a `Set<string>` getter | `StackEntry` class linked list |
| `doResolve` hot path | direct POJO alloc + `hasStackEntry` field-compare | `instanceof Set` branch + possible `setToStack` conversion + eager string build | direct `new StackEntry(..., parent)` + field-compare `has` |
| Per-call string formatting | lazy (only on recursion log) | every call | lazy (only on recursion log) |
| `stack.has(entry)` (public API) | **breaks** — stack is a POJO, no `.has` | works (but materializes a Set each read) | works (O(n) list walk, no alloc) |
| `stack instanceof Set` | **breaks** | true | **breaks** — rare in practice |
| `stack.add(x)` | **breaks** | works (on the Set view) | **breaks** — external code doesn't mutate |
| Webpack `ResolverCachePlugin` | **requires update** (author's own note) | works (via Set conversion path) | works (no change needed) |

### CodSpeed CI benchmarks (vs `main` BASE)

| Benchmark | PR #523 | **this PR (#526)** |
|---|---|---|
| pathological-deep-stack: alias chain of 50 (warm) | +50.91% | +48.65% |
| realistic-midsize: mixed batch (warm cache) | +10.29% | **+12.55%** |
| resolve-to-context: directory resolve (warm) | +11.82% | **+18.65%** |
| symlinks: symlinks=false (warm) | +10.77% | **+12.38%** |
| description-files-multi (warm) | +11.07% | **+13.13%** |
| exports-patterns-many: 6 prefixes × 4 leaves (warm) | +10.16% | **+11.2%** |
| main-field: browser/module/main combos (warm) | +13.68% | *(similar)* |
| **cache-predicate: mixed cached/uncached (warm)** | **−45.02% ❌** | **+10.5% ✅** |

PR #523 regresses `cache-predicate` by 45% because its `doResolve` hot path adds an `instanceof Set` branch + `_stack \|\| stack` fallback + lazy Set getter materialization. For paths that hit UnsafeCachePlugin and skip the rest of the plugin chain (already microsecond-scale), that per-call overhead dominates.

This PR avoids the regression by keeping a single representation (the linked list) with no branching or materialization on the hot path — every benchmark improves, none regress.

perf

**Use of AI**

### DIFF
--- a/.changeset/perf-stack-entry-linked-list.md
+++ b/.changeset/perf-stack-entry-linked-list.md
@@ -1,0 +1,14 @@
+---
+"enhanced-resolve": patch
+---
+
+Replace the `Set<string>`-based resolver stack with a singly-linked `StackEntry` class that exposes a Set-compatible API (`has`, iteration, `size`, `toString`).
+
+Each `doResolve` call now prepends a single linked-list node instead of cloning the entire Set, making stack push O(1) in time and memory. Recursion detection walks the linked list (O(n)), but because the stack is typically shallow this is much cheaper than cloning a Set per call.
+
+Because `StackEntry` implements the Set methods consumers use in practice, `resolveContext.stack` remains a drop-in replacement — existing callers that iterate or call `.has()` keep working.
+
+Benchmarks on the local suite:
+
+- `pathological-deep-stack` (50-level alias chain): +57%
+- `realistic-midsize` warm cache: +16%
--- a/lib/Resolver.js
+++ b/lib/Resolver.js
@@ -366,11 +366,6 @@ const _pathCacheByFs = new WeakMap();
 /** @typedef {BaseResolveRequest & Partial<ParsedIdentifier>} ResolveRequest */
 
 /**
- * String with special formatting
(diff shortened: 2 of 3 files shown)


alexander-akait (approved):
Let's merge, I don't think anyone uses `add` method for stack, perf improvements are looking very good,

PR #564, opened 2026-05-03
fix: recognize win32 relative paths in getType and isRelativeRequest

### LINKED ISSUE #401: Add ability to handle win32 relative paths

xiaoxiaojx:
**Summary**
fixes https://github.com/webpack/enhanced-resolve/pull/436

fix

Yes

No

Nothing


alexander-akait (commented):
To be honestly I am not sure we really need to allow this, we should check is `require.resolve` and `import.meta.resolve` supports win32 paths, if yes, we will ship it, if no, it is limitation, like `globs`, they are support only `/`

PR #626, opened 2026-07-09
feat(profiler): add resolution performance profiler plugin

hey-amanthakur:
Implements the ProfilerPlugin that taps the resolveStep sync hook to record per-hook entry timestamps, maintains a frame stack for duration tracking, and delivers profile data via resolveContext.profile callback or factory-level profileCallback.

The \`profile\` option is added to ResolverFactory.createResolver options:
- \`profile: true\` enables profiling with per-call resolveContext.profile
- \`profile: (data) => {}\` enables profiling with a factory-level callback
- profile is propagated through createInnerContext for nested resolves

Changes:
- lib/ProfilerPlugin.js: new plugin with frame-stack timing
- lib/ResolverFactory.js: profile option + wiring
- lib/Resolver.js: profile field in ResolveContext typedef
- lib/createInnerContext.js: propagate profile to inner context
- lib/index.js: export ProfilerPlugin via lazy getter
- types.d.ts: ProfileData, ProfileHookEntry, ProfilerPlugin types
- test/profiler.test.js: 6 tests covering all delivery paths

### DIFF
--- a/lib/ProfilerPlugin.js
+++ b/lib/ProfilerPlugin.js
@@ -1,0 +1,178 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author jhonsnow456
+*/
+
+"use strict";
+
+/** @typedef {import("./Resolver")} Resolver */
+/** @typedef {import("./Resolver").ResolveStepHook} ResolveStepHook */
+/** @typedef {import("./Resolver").ResolveRequest} ResolveRequest */
+/** @typedef {import("./Resolver").ResolveContext} ResolveContext */
+
+/**
+ * @typedef {object} ProfileHookEntry
+ * @property {string} name hook name
+ * @property {number} count number of times this hook was entered
+ * @property {number} totalTime total wall-clock time in milliseconds
+ */
+
+/**
+ * @typedef {object} ProfileData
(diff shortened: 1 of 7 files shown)


alexander-akait:
Why? Is it AI generated?

hey-amanthakur:
@alexander-akait 

When debugging slow or failing resolves, I currently have two options: resolveContext.log (unstructured strings, no filtering) or tapping resolveStep myself (which only tells you what was entered, not how long it took or whether it bailed). Neither gives me per-hook timing data for a single resolve call , I need that to pinpoint which step is the bottleneck on large builds with deep dependency trees.


No,  the frame-stack approach in ProfilerPlugin.js is modeled on how Chrome DevTools records call stacks (push on entry, pop on exit, attribute elapsed time to parent). The implementation choices, tapping resolveStep/result/noResolve as the three lifecycle points, propagating profile through createInnerContext the same way log is propagated, and the profile: true | callback dual API : follow the same patterns already used by log and resolveContext in Resolver.js. The tests use memfs to get deterministic file system behavior and verify all three delivery paths (resolveContext.profile, factory callback, and log output).

alexander-akait:
You can profile a code to see results, don’t think we need a plugin for this, also this is AI generated code

alexander-akait:
Feel free to feedback

PR #627, opened 2026-07-10
feat: add TracePlugin for structured resolution tracing

hey-amanthakur:
Add a TracePlugin that hooks into the resolver lifecycle to record structured trace entries for every resolution step. Each entry captures the hook name, request details, result, and timestamp as a machine-readable object instead of plain text strings.

This allows developers to filter, search, and programmatically analyze resolution traces, complementing the existing resolveContext.log which outputs unstructured text.

### DIFF
--- a/lib/TracePlugin.js
+++ b/lib/TracePlugin.js
@@ -1,0 +1,135 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Aman Thakur @jhonsnow456
+*/
+
+"use strict";
+
+/** @typedef {import("./Resolver")} Resolver */
+/** @typedef {import("./Resolver").ResolveRequest} ResolveRequest */
+
+/**
+ * @typedef {object} TraceRequest
+ * @property {string} path resolving directory
+ * @property {string} request request string
+ * @property {string} query query string
+ * @property {string} fragment fragment string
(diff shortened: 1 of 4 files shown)


alexander-akait:
Describe how it helps you, thanks

hey-amanthakur:
@alexander-akait 

Right now, when a resolve fails or takes unexpectedly long, my only option is resolveContext.log which emits unstructured strings. To debug I have to pass { log: console.log } and manually parse the indented output — there's no way to filter by pipeline step, grep for failed candidates, or feed the trace into tooling.

A structured TraceEntry[] fixes this for my use case:
- Filtering: I can entries.filter(e => e.hook === "resolve-as-module") instead of regex-matching strings.
- Failure diagnosis: On a failed resolve, I get the full path taken through the pipeline with timestamps, rather than error.details which is just log.join("\n").
- No double-resolve overhead: When resolve() fails without a log callback, the resolver re-runs the entire pipeline just to populate error.details. A pre-built trace buffer makes that redundant.
- Tooling: I can feed the trace array into a visualizer or diff it across builds, which is the main motivation behind the PR title.

alexander-akait:
We don’t need this into core, you can use a custom plugin if you need it, there are literally couple lines

PR #628, opened 2026-07-10
test(alias): add unit tests for compileAliasOptions bucketing and onl…

hey-amanthakur:
- compileAliasOptions: verify bucket construction by first char code, useBuckets logic (disabled for single-char or empty-prefix wildcard), wildcard prefix/suffix splitting, absolutePath, and empty input

- onlyModule flag: verify relative paths and path-prefixed requests are skipped when onlyModule is true, and that both are aliased when false


alexander-akait:
Lint failed

hey-amanthakur:
@alexander-akait have fixed the linting issue, It could be due to new type.d.ts has not been generated.

PR #639, opened 2026-07-27
fix(RestrictionsPlugin): treat only the platform separator as a boundary

stormslowly:
## Why

`isInside` accepted both `/` and `\` as the segment boundary after a string restriction, unconditionally. On posix a backslash is a regular filename character, so a file that merely *starts* with the restriction string escapes it.

With `restrictions: ["/a/b/c"]`:

| request | before | after |
| --- | --- | --- |
| `/a/b/c/index.js` | allowed | allowed |
| `/a/b/c\sibling.js` (a file named `c\sibling.js` in `/a/b`) | **allowed** | rejected |

The escaping file is a sibling of the restricted directory, not inside it.

## What

The boundary is now compared against `path.sep` only, so a backslash separates segments on Windows and is part of the filename everywhere else. This matches the `Path::starts_with` semantics [rspack_resolver](https://github.com/web-infra-dev/rspack-resolver) uses for the same check.

`lib/util/path-browser.js` gains `sep` — the shim only exposed the subset `lib/util/path.js` needed, and without it `sep` would be `undefined` in browser bundles. A browser has no host platform and its paths are posix-like, so it is `/`.

Tests are per-platform: the escape case runs on posix, the "a backslash still separates" case runs on Windows.

### DIFF
--- a/.changeset/restrictions-path-separator-escape.md
+++ b/.changeset/restrictions-path-separator-escape.md
@@ -1,0 +1,5 @@
+---
+"enhanced-resolve": patch
+---
+
+Stop a filename containing a backslash from escaping a string `restrictions` entry on posix, by only accepting the platform separator as a segment boundary.
--- a/lib/RestrictionsPlugin.js
+++ b/lib/RestrictionsPlugin.js
@@ -5,11 +5,12 @@
 
 "use strict";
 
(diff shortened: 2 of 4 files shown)


alexander-akait:
Thanks for the PR, fixed in the main with some changes

PR #640, opened 2026-07-27
fix(RestrictionsPlugin): accept a restriction ending with a separator

stormslowly:
> Stacked on #639 — its commit is included here because a cross-fork PR cannot use another fork branch as its base. Only the second commit belongs to this PR.

## Why

A string restriction that ends with a separator rejects everything inside it.

`isInside` checks the character *after* the restriction for a separator, which assumes the restriction does not end with one. With `restrictions: ["/a/b/c/"]` and the path `/a/b/c/index.js`, that character is `i` — the first character of the filename — so the file is reported as outside the restriction and resolution fails.

| restriction | request | before | after |
| --- | --- | --- | --- |
| `/a/b/c` | `/a/b/c/index.js` | allowed | allowed |
| `/a/b/c/` | `/a/b/c/index.js` | **rejected** | allowed |

A restriction already ending with a separator carries its own boundary, so `startsWith` is enough.

### DIFF
--- a/.changeset/restrictions-path-separator-escape.md
+++ b/.changeset/restrictions-path-separator-escape.md
@@ -1,0 +1,5 @@
+---
+"enhanced-resolve": patch
+---
+
+Stop a filename containing a backslash from escaping a string `restrictions` entry on posix, by only accepting the platform separator as a segment boundary.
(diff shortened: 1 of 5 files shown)


alexander-akait:
Thanks for the PR, fixed in the main with some changes

PR #641, opened 2026-07-28
fix: keep the original request resolvable when extensionAlias lists its own extension

alexander-akait:
**Summary**

`ExtensionAliasPlugin` marks every candidate `fullySpecified: true` and then refuses to fall back, so a request ending in an aliased extension can only ever resolve as an exact file — never as a directory or via `package.json#main`. With the source extension listed among its own aliases (`{ ".js": [".js", ".ts"] }` — the documented TypeScript recipe, and what `experiments.typescript` sets in webpack) that entry is meant to keep the original request valid, but it is tried fully specified too, so any package or directory whose name ends in `.js` becomes unreachable. This makes the identity entry mean what it says: after the fully specified candidates fail, fall through to normal resolution. A mapping that drops the source extension (`{ ".js": [".ts"] }`, `{ ".mjs": ".mts" }`) stays strict.

Refs webpack/webpack#21541. That regression is fixed on the webpack side in webpack/webpack#21542 and does not need this; this closes the same failure for projects that genuinely enable TypeScript today, and ahead of TypeScript becoming a default.

fix

Yes — three cases in `test/extension-alias.test.js` (directory named `dir3.js`, an aliased absolute path to a `pkg.js` package, and a strict mapping that must keep failing) plus the matching fixtures.

No — it only turns previously failing resolutions into successes; exact-file candidates are still tried first, so nothing that resolves today resolves differently.

`extensionAlias` docs should say that listing an extension among its own aliases keeps the original request resolvable in its normal form, while omitting it makes the mapping strict.

**Use of AI**

AI (Claude Code) was used to trace the failure, draft the fix and tests, and verify them; the diagnosis and the final diff were reviewed by a human before submitting.

### DIFF
--- a/.changeset/010-extension-alias-original-request-fallback.md
+++ b/.changeset/010-extension-alias-original-request-fallback.md
@@ -1,0 +1,5 @@
+---
+"enhanced-resolve": patch
+---
+
+Keep the original request resolvable when `extensionAlias` lists its own extension.
--- a/lib/ExtensionAliasPlugin.js
+++ b/lib/ExtensionAliasPlugin.js
@@ -114,6 +114,14 @@ module.exports = class ExtensionAliasPlugin {
 				const stoppingCallback = (err, result) => {
 					if (err) return callback(err);
 					if (result) return callback(null, result);
+					// Listing the source extension among its own aliases keeps the
+					// original request valid, so it must still be resolvable in its
+					// normal (not fully specified) form: as a directory, or with
+					// extensions appended. Only a mapping that drops the source
+					// extension is strict.
+					if (isAliasString ? alias === extension : alias.includes(extension)) {
+						return callback();
+					}
 					// Don't allow other aliasing or raw request
 					return callback(null, null);
 				};
--- a/test/extension-alias.test.js
+++ b/test/extension-alias.test.js
@@ -67,6 +67,52 @@ describe("extension-alias", () => {
 		});
 	});
 
+	describe("fallback to the original request", () => {
+		it("should resolve a directory whose name ends with the aliased extension", (t, done) => {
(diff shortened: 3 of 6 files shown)


alexander-akait:
Closing — this was the wrong layer.

`ExtensionAliasPlugin`'s strict no-fallback behavior is not a regression: it is byte-identical in 5.12.0 (when `extensionAlias` was introduced), 5.16.0 and 5.24.3, and it is asserted on purpose by `"should not allow to fallback to the original extension or add extensions"`.

The actual cause of webpack/webpack#21541 is on the webpack side: `experiments.typescript: "auto"` checks only that Node.js >= 22.6 can strip types and that no TS loader is registered — never whether the project contains any TypeScript — so every project on a recent Node.js started getting TypeScript resolution semantics, including this `extensionAlias`. Fixed in webpack/webpack#21542, which needs no change here.

---

alexander-akait:
Reopened with a corrected rationale (title and description updated); my earlier comment was right that this is not the cause of webpack/webpack#21541, but wrong to conclude nothing needed fixing here.

The webpack-side fix (webpack/webpack#21542) stops `experiments.typescript: "auto"` from applying this `extensionAlias` to projects that never asked for TypeScript. It does not help when TypeScript is genuinely enabled — verified against that fix, `resolve.alias` pointing at a package directory named `tippy.js` still fails under `experiments.typescript: true` and under `futureDefaults` (webpack 6's defaults). Isolating the cause with webpack's own generated resolve options: dropping `tsconfig` still fails, dropping `extensionAlias` resolves. So this is the piece that keeps a package or directory whose name ends in `.js` unreachable, today for explicit opt-in and by default once TypeScript stops being opt-in.

---

PR #676, opened 2026-09-21
feat: explain exports/imports conditions that wrap subpaths

### LINKED ISSUE #325: Exports field trees with top-level conditions result in a misleading error message
When the exports field of a package has top-level conditions leading to subpath conditions, enhanced-resolve does not parse the subpath conditions. Following a discussion with the maintainer, this is expected behaviour as that syntax is not supported by Node.js either.

However, enhanced-resolve parses the exports field without failing, and later on in a Webpack compilation, Webpack fails to resolve subpath imports. Instead, enhanced-resolve should detect this syntax and throw an error message somewhere around `buildExportsFieldPathTree`. The reasoning is that it's very hard to understand why subpath imports fail otherwise, because the Webpack and Node.js documentations don't explicitly state that this use of exports is unsupported.

## Original ticket
(original title: Exports field trees with top-level conditions are not supported)

From my understanding, enhanced-resolve is supposed to know how to handle the `exports` field of `package.json`, including exports with conditional names, as there are checks for conditional mappings in the code.

Everything in this ticket relates to the `lib/util/entrypoints.js` file.

In the `buildExportsFieldPathTree`, we can see that exports that do not start with a dot are rejected. Conditional mappings are checked _after_ the building of the `treeRoot`, so this means that exports with top-level conditions are wrongly assumed to be incorrect syntax. This results in `treeRoot` containing a single file entry that matches only `.` and has the entire `exports` field as its content.

package.json:
```json
	"exports": {
		"import": {
			".": "./esm/index.js",
			"./*": "./esm/*.js"
		},
		"require": "./build/bundle.js"
	},
```

Webpack.config.js:
```js
  config.resolve.conditionNames = ['import', 'node', 'default']
```

Obtained `treeRoot`:
```js
{
  children: null,
  folder: null,
  wildcards: null,
  files: Map(1) { '' => { import: { ".": "./esm/index.js", "./*": "./esm/*.js" }, require: './build/bundle.js' }
}
```

As a consequence, imports of the form `@org/mylib` work, but `@org/mylib/foo` fail. Should you need a reproduction example, the one written in the Webpack 5 documentation does not work: https://webpack.js.org/guides/package-exports/#providing-commonjs-and-esm-version-stateful.

I am not sure what the appropriate approach is, hence the lack of patch.

One could perform a conditional mapping match before building `treeRoot`, and building the first matching sub-tree, resulting in a `treeRoot` like so:
```js
{
  children: null,
  folder: null,
  wildcards: Map(1) { '' => './esm/*.js' },
  files: Map(1) { '' => './esm/index.js' }
}
```

Or one could map conditional trees into trees with a top-level relative path and conditional leaves, resulting in something like:
```js
{
  children: null,
  folder: null,
  wildcards: Map(1) { '' => './esm/*.js' },
  files: Map(1) { '' => { import: "./esm/index.js", "require": "./build/bundle.js" } }
}
```

Or the treeRoot structure itself could be recursive, though I've no idea what impact this would have on the rest of the code.

alexander-akait:
**Summary**

An `exports` field shaped like `{ "import": { ".": "./esm/index.js", "./*": "./esm/*.js" }, "require": "./build/bundle.js" }` is not supported by Node.js: a condition may only map to a target or to further conditions, so the subpaths under `"import"` are read as condition names and match nothing. The failure surfaced as `"./foo" is not exported under the conditions [...]`, which points at the request rather than the mistake and reads as though the package forgot to export `./foo` — #325 took its reporter a day to diagnose. That error now carries the reason: the offending keys, and the arrangement that works with the subpaths at the top level and the conditions nested inside them. Found at any depth and inside array targets, with the same diagnosis for `#` keys under a condition in `imports`. Closes #325.

feat — a better error message; resolution behaviour is unchanged.

Yes — `test/entrypoints.test.js` covers the shorthand form, a nested offender, an offender inside an array target, the `imports` counterpart, and the cases that must stay silent. The 1616 existing tests are unchanged and pass.

No, and deliberately so. The diagnosis runs only where a request has already failed, never while building the field, so nothing that resolves today starts failing — the field above still resolves `.` under `require`. Validating the field instead would have broken that, leaving a consumer who cannot patch the upstream package with no way forward, and it would have cost a full tree walk of every `package.json` to catch a mistake almost none of them make: on a 1000-subpath field that took construction from 0.0674 ms to 0.2888 ms. As written, construction stays at main's cost (0.0648 vs 0.0674 ms, retained heap identical at 2.83 KB) and resolve throughput is unmoved — interleaved runs of the `exports` benchmarks gave 9089 and 9217 ops/s against main's 8811 and 9141, a gap smaller than the spread within either arm.

n/a — the changeset covers the changelog.

**Use of AI**

Written with Claude Code. It measured rather than assumed at each step: it reproduced the current behaviour first (under `import` the field resolves nothing at all, not just subpaths), instrumented a benchmark run to confirm the field processor is built once across 3823 samples, and benchmarked an eager-validation version against this one, which is what moved the design from rejecting the field to diagnosing the failure. Reviewed by me before opening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF

---
_Generated by [Claude Code](https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resolution errors for unsupported `exports` and `imports` configurations by identifying problematic subpaths and showing the supported structure.
  - Errors while processing these fields now identify the relevant `package.json` file for easier troubleshooting.
  - Resolution behavior remains unchanged; successful resolutions are unaffected.

- **Tests**
  - Added coverage for nested, conditional subpath diagnostics across `exports` and `imports` configurations, including valid and invalid field formats.


alexander-akait:
**CodSpeed** is red ("Performance Regression: -7.81%"). This PR does touch `lib/`, so that deserved checking rather than waving away — but the one regressed benchmark cannot be affected by any change to this package.

The flagged case is `node-compare: node require.resolve x 1000` (Memory), and its body is:

```js
bench.add(`node-compare: node require.resolve x ${BATCH_SIZE}`, () => {
	for (let i = 0; i < requests.length; i++) {
		requireAnchor.resolve(requests[i]);
	}
});
```

`requireAnchor` is `createRequire(path.join(srcDir, "index.js"))`. That task measures **Node's own CJS resolver** — it is the control arm of the head-to-head case and never calls enhanced-resolve. No edit to `lib/` can move it.

Two further points against reading it as real:

- The same benchmark's `BASE` was **78.7 KB** on #675 (base `332d999`) and is **59.3 KB** here (base `e1bdb75`). Node's `require.resolve` memory moving 25% between two base commits of *our* package, neither of which can affect it, is the instrument.
- The other flagged row is an **improvement** — `realistic-midsize: mixed batch (cold cache)`, 18.1 ms to 15.5 ms, +16.98%. My diff cannot make an enhanced-resolve benchmark 17% faster either.

Independently of the benchmark: the diagnosis this PR adds runs only where `paths.length === 0`, i.e. after a request has already failed to match. No successful resolve reaches it. Measured locally, field construction is unchanged against `main` (0.0648 vs 0.0674 ms on a 1000-subpath field, retained heap identical at 2.83 KB), and interleaved runs of the `exports` benchmarks gave 9089 and 9217 ops/s on this branch against 8811 and 9141 on `main` — a gap smaller than the spread within either arm.

So there is nothing to fix and no fix to port. I am re-running the benchmark workflow once; if it comes back red on the same warning, it is measurement noise rather than this diff. Acknowledging it in the CodSpeed dashboard is a maintainer action, so I am leaving that to you.

---

PR #679, opened 2026-09-21
test(alias): cover compileAliasOptions bucketing and the onlyModule flag

alexander-akait:
**Summary**

Picks up @hey-amanthakur's #628, which went stale against `main` (conflicted, lint red) — original commit kept as co-author. Closes #628.

`compileAliasOptions` builds the structure every resolve reads: the first-char buckets, the `useBuckets` switch that decides whether the hot path pays for a `Map.get`, and the per-option precomputed strings (`nameWithSlash`, `absolutePath`, `wildcardPrefix`/`wildcardSuffix`, `firstCharCode`, `arrayAlias`). Both bucket layouts resolve identically, so no resolve-level test can tell them apart — a regression in the layout, or in the conditions that turn bucketing off, is invisible until it silently stops aliasing something. These assert the compiled shape directly.

Worth being precise about what this buys: `lib/AliasUtils.js` is already at 100% line coverage on `main` (333/333), so this is not a coverage patch. What it adds is behavioural pinning that line coverage does not give — declaration order inside a bucket (alias precedence depends on it), `useBuckets` off for a single first char and for an empty-prefix wildcard, the any-first-char entry (`firstCharCode === -1`) being kept out of the buckets, and the fact that a windows-style name normalizes `absolutePath` with a trailing `\` while `nameWithSlash` appends `/`, which is exactly why `aliasResolveHandler` compares against both.

Changes from #628 as submitted:

- Rebased onto current `main`; #628 also carried a `types.d.ts` reformat that an older Prettier had produced. That file is now a two-line stub over generated `types/` (#675), so the churn is gone rather than reapplied.
- The `compileAliasOptions` tests used a `{ join: (filePath, ext) => filePath + ext }` stub, which made `absolutePath` come out as `/abs/path` and the test pinned that. The resolver builds it as `join(name, "_").slice(0, -1)`, so the real value is `/abs/path/` — the stub was pinning a shape the resolver never produces. These use a real resolver instead.
- Added: declaration order within a bucket, a two-wildcard name (not treated as a wildcard alias), the windows `absolutePath` form, `arrayAlias`, and the `onlyModule` normalization. The `onlyModule` resolve tests now build one resolver per case through a helper rather than repeating the fixture three times, and cover a subpath (`real/sub/file`) as well as a relative request.

test — no `lib/` change, so resolution behaviour and performance are untouched.

That is the change: 13 tests in `test/alias.test.js`. The suite is 1629 tests, all passing, and `npm run lint` is clean across all five stages.

No.

n/a — test-only, so no changeset either (nothing published changes).

**Use of AI**

Written with Claude Code, on top of @hey-amanthakur's original commit. It checked the assertions against the code rather than carrying them over: it measured `lib/AliasUtils.js` at 333/333 lines on `main` both with and without these tests, which is why the summary above does not claim a coverage gain; it ran `compileAliasOptions` with a real resolver to find that the submitted `absolutePath` assertion pinned the stub's output rather than the resolver's; and it confirmed the windows-name case is host-independent (`join` dispatches to `winNormalize` on the path type, not on the platform) before relying on it off Windows. Reviewed by me before opening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF

---
_Generated by [Claude Code](https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF)_

## Summary by CodeRabbit

* **Tests**
  * Added coverage for alias configuration handling, including wildcard patterns, path normalization, and option flags.
  * Added tests verifying exact-match-only aliasing and subpath resolution behavior.


alexander-akait:
**CodSpeed** is red on `85d83be` ("Performance Regression: -10.05%"). It is not this PR's, and the report says why in its own footer: it compared `85d83be` against **`main` (f07a030)**, which is #676 — merged ~10 minutes before this branch was pushed. The branch was cut from `d9a13d7`, so the arms differed by #676's `lib/` changes, not by anything here.

The diff is one file:

```
 test/alias.test.js | 236 +++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 236 insertions(+)
```

No `lib/` change at all, and the flagged benchmark imports only `../../../lib/index.js`, `fs` and `path` — nothing in the benchmark harness loads `test/`. A test file cannot move a benchmark.

The mirror-image confirms the base skew rather than a real delta. `realistic-midsize: mixed batch (cold cache)` was reported on #676 as an **improvement**, 18.1 ms → 15.4 ms (+17.34%). Here, with the arms swapped, the same task is reported as a **regression**, 15.7 ms → 18.2 ms (-13.69%). Same pair of readings, opposite sign, because this branch was the side without #676.

Rather than argue it from the report, I have merged `main` into the branch (`f4dfa6d`). `git diff HEAD origin/main -- lib/` is now empty, so the next CodSpeed run compares identical `lib/` on both sides and anything it still reports is instrument noise. Locally after the merge: 1635 tests pass and `npm run lint` is clean across all five stages.

I have not spent a manual re-run — the push re-runs the benchmarks on an aligned base, which is the better measurement. Acknowledging anything in the CodSpeed dashboard is a maintainer action, so I am leaving that to you.

---

alexander-akait:
Correcting my previous comment: aligning the base did **not** clear CodSpeed. It is still red on `f4dfa6d` (-9.27%, same two benchmarks). My prediction there was wrong — but the result is more conclusive than the base-skew explanation was, not less.

After the merge, `f4dfa6d` differs from `main` (`f07a030`) by exactly this:

```
$ git diff f07a030 f4dfa6d --stat
 test/alias.test.js | 236 +++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 236 insertions(+)

$ git diff f07a030 f4dfa6d -- lib/ benchmark/ package.json package-lock.json | wc -l
0
```

`lib/`, `benchmark/`, `package.json` and the lockfile are byte-identical on the two sides. CodSpeed is measuring the same code against itself and reporting a 9.27% difference, so the difference is the instrument, not the code. The report says as much in its own header: *"Different runtime environments detected — some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results."*

`alias-wildcard-scan: 100+1 wildcard + 1 exact` in particular reports **1.5 KB → 2.7 KB, -46.84%** — the identical figures it gave for `85d83be`, a different commit. A reading that does not move when the commit does is pinned to the environment, not to the tree.

I have now spent the one re-run (Benchmarks run 35573756485). Whatever it returns, there is nothing in this PR to fix: there is no `lib/` change to revert, tune or bisect. Acknowledging it in the CodSpeed dashboard is a maintainer action, so I am leaving that to you, and I will keep watching the PR until it merges or closes.

---

